Issue #10, opened 2019-08-13
Setup views 


songz:
create views folder and setup ejs rendering template

songz:
Done.

PR #17, opened 2019-08-14
#10 Creates views folder/template for Admin ServiceKeys UI

SahilKalra98:


src/app.ts
```diff
@@ -1,8 +1,11 @@
 import express from 'express'
-
+import path from 'path'
 const app = express()
 const port = 3000
 
-app.get('/', (req, res) => res.send('Hello World!'))
+app.set('view engine', 'ejs')
+app.set('views', path.join(__dirname, '../../views'))
```
songz: could we have `views` inside the `src` folder please? I think its better to keep the root folder clear.

Issue #8, opened 2019-08-13
create setup script


songz:
when users run `yarn setup`, the following should happen:

1. clone [this project](https://github.com/Neilpang/acme.sh) into project root (should be in gitignore)
2. run `chmod +x ./acme/acme.sh` so that the script has execution access
3. run `yarn`

Wolfy64:
#19 [Create a setup Script](https://github.com/garageScript/myProxy/pull/19)

PR #19, opened 2019-08-15
Create a setup Script

Wolfy64:
#8 Create a setup Script


script.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
songz: I would imagine in the future we will have more scripts. could we create and put this in a `scripts` folder? maybe `scripts/setup.sh`

Issue #9, opened 2019-08-13
Admin page setup


songz:
Admin app should run on port `20000`.

if no cookie, prompt for admin password. Admin password input = access to admin page.

Admin password is set up by environment variables. For example:

`ADMIN=9458237 yarn start`

PR #25, opened 2019-08-17
#9 Admin Page Login Setup

SahilKalra98:


src/app.ts
```diff
@@ -1,13 +1,24 @@
...
+app.get('/login', (req, res) => res.render('login', { error: '' }))
+app.post('/login', (req, res) => {
+  if (process.env.ADMIN !== req.body.adminPass)
+    return res.render('login', { error: 'Wrong Admin Password' })
+  res.cookie('adminPass', req.body.adminPass)
+  res.redirect('/admin/serviceHostKeys')
+})
 app.get('/', (req, res) => res.render('index', { message: 'Hello myProxy' }))
 
 app.get('/admin/serviceHostKeys', (req, res) => {
```
songz: You will need a check for all admin routes, so you need to add a middleware so all requests going through admin will get a check. Sample:

```
app.use('/admin', (req, res, next) => {
  if (!req.cookies.adminPass) return res.redirect('/login')
  return next()
})
```

Running the next function will continue the request to other handlers.

src/app.ts
```diff
@@ -1,13 +1,30 @@
 import express from 'express'
 import path from 'path'
+import cookieParser from 'cookie-parser'
 const app = express()
 const port = process.env.PORT || 3000
+const adminRouter = express.Router()
```
songz: create a new folder called `admin`. Then create a file inside `admin` called `index.ts`. Then put this router stuff inside that file and export it. Let's not clutter the main file.

src/admin/index.ts
```diff
@@ -0,0 +1,13 @@
+import express from 'express'
+const adminRouter = express.Router()
+
+adminRouter.use('/*', (req, res, next) =>  {
+  if (!req.cookies.adminPass) return res.redirect('/login')
+  next()
+})
+
+adminRouter.get('/serviceHostKeys', (req, res) => {
+  res.render('admin')
```
songz: I would create a folder called `admin`, and then call this view `admin/serviceHostKeys`

PR #32, opened 2019-08-23
WIP: Admin UI functionality

SahilKalra98:


src/app.ts
```diff
@@ -4,6 +4,7 @@ const app = express()
 const port = process.env.PORT || 3000
 app.set('view engine', 'ejs')
 app.set('views', path.join(__dirname, '../../src/views'))
+app.use(express.static(`${__dirname}`))
```
songz: woah... this makes the entire app publicly accessible, so now everybody can see the server code / database.
Let's not do this. Stick to `public` folder.

Issue #11, opened 2019-08-13
Admin API to manage api key / secret


songz:
Create REST apis for `serviceHostKeys` to store keys to be used for [acme ssh key generation](https://github.com/Neilpang/acme.sh/wiki/dnsapi#18-use-gandi-livedns-api)

### POST `/api/admin/serviceHostKeys`
```
{
  key: 'DNSimple_OAUTH_TOKEN',
  value: 'User-Input',
  service: 'dns_dnsimple'
}
```

PATCH, PUT, DELETE, GET should be self explanatory.

API will be used to power the UI here: https://github.com/garageScript/myproxy/issues/12

songz:
Refer to image: https://github.com/garageScript/myproxy/issues/12

PR #35, opened 2019-08-25
 Follows up on serviceHost API #11

SahilKalra98:


Wolfy64:
I did a PR to fixed Merge conflict and fix the path view when we `run build:live`

src/api/admin.ts
```diff
@@ -0,0 +1,61 @@
+import { DB, ServiceKey } from './types/admin'
+import { setData, getData, data } from './lib/data'
+import express from 'express'
+import fs from 'fs'
```
songz: fs is not used in this file.

src/api/admin.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const app = express.Router()
+
+const serviceKeys = getData('serviceKeys')
+
+app.post('/serviceHostsKey', (req, res) => {
+  // create service keys
+  const serviceHostKeys: ServiceKey = {
+    id: data.serviceKeys.length,
```
songz: I'm not sure if using the length is a good idea, because if a service key gets deleted, then the id could see a duplicate. For example:
4 keys are created. So the id will be: `0,1,2,3`

Then one key (`id: 1`) is deleted, so ids will be: `0,2,3`

If another key is created after that, then the ids will be `0,2,3,3`. This is a potential future bug. 

The I would generate ids via `uuid v4`: https://www.npmjs.com/package/uuid

src/api/admin.ts
```diff
@@ -0,0 +1,61 @@
...
+    ...req.body
+  }
+  data.serviceKeys.push(serviceHostKeys)
```
songz: should not be `data.serviceKeys`. Should be just `serviceKeys`.

src/api/admin.ts
```diff
@@ -0,0 +1,61 @@
+import { DB, ServiceKey } from './types/admin'
+import { setData, getData, data } from './lib/data'
```
songz: you don't need to import data.

src/api/admin.ts
```diff
@@ -0,0 +1,61 @@
...
+import express from 'express'
+import fs from 'fs'
+
+const app = express.Router()
+
+const serviceKeys = getData('serviceKeys')
```
songz: remove, serviceKeys might not be ready yet when the app first runs. You should get the serviceKeys on every request.

src/api/admin.ts
```diff
@@ -0,0 +1,61 @@
...
+
+app.post('/serviceHostsKey', (req, res) => {
+  // create service keys
```
songz:
```
const serviceKeys = getDate('serviceKeys')
serviceKeys.push( {....})
setData('serviceKeys', serviceKeys)
```

src/api/admin.ts
```diff
@@ -0,0 +1,61 @@
...
+    ...req.body
+  }
+  data.serviceKeys.push(serviceHostKeys)
+  setData('servicekeys', data)
+  res.json(serviceHostKeys)
+})
+
+app.get('/serviceHostsKey', (req, res) => {
+  // get all servicekeys
+  res.json(data.serviceKeys)
```
songz:
```
const serviceKeys = getData('serviceKeys')
res.json(serviceKeys)
```

src/api/admin.ts
```diff
@@ -0,0 +1,61 @@
...
+})
+
+app.get('/serviceHostsKey/:id', (req, res) => {
+  // grab one servicekey
+  const id: number = Number(req.params.id)
+  const selectedKey: ServiceKey = data.serviceKeys[id]
```
songz: use a filter.

src/api/admin.ts
```diff
@@ -0,0 +1,61 @@
...
+  res.json(selectedKey)
+})
+
+app.delete('/serviceHostsKey/:id', (req, res) => {
+  // delete a servicekey
+  const id: number = Number(req.params.id)
+  data.serviceKeys.splice(id, 1)
```
songz: filter to get the new array, then save data

src/api/lib/data.ts
```diff
@@ -0,0 +1,32 @@
...
+  const fileData: string = JSON.stringify(data)
+  fs.writeFile('./data.db', fileData, err => {
+    if (err) {
+      return console.log('writing to DB failed', err)
+    }
+    console.log('successfully wrote to DB')
+  })
+}
+
+export { getData, setData, data }
```
songz: no need to export data

src/api/admin.ts
```diff
@@ -0,0 +1,80 @@
...
+  // get all servicekeys
+  const serviceKeys = getData('serviceKeys')
+  res.json(serviceKeys)
+})
+
+app.get('/serviceHostsKey/:id', (req, res) => {
+  // grab one servicekey
+  const serviceKeys = getData('serviceKeys')
+  const selectedKey = serviceKeys.filter(
+    element => element.id === Number(req.params.id)
```
songz: `uuid4()` returns a string. If you convert `req.params.id`, you will never find a match.

songz: not sure why you are converting all the input params into number, but please correct for all API endpoints.

src/api/admin.ts
```diff
@@ -0,0 +1,80 @@
...
+})
+
+app.delete('/serviceHostsKey/:id', (req, res) => {
+  // delete a servicekey
+  let deletedKey
+  const serviceKeys = getData('serviceKeys')
+  const updatedKeys = serviceKeys.filter(element => {
+    if (element.id !== Number(req.params.id)) return true
+    if (element.id === Number(req.params.id)) deletedKey = element
+    return false
```
songz: `return element.id !== req.params.id`

src/api/admin.ts
```diff
@@ -0,0 +1,80 @@
...
+  let editedKey
+  const serviceKeys = getData('serviceKeys')
+  const id: number = Number(req.params.id)
+  const updatedKeys = serviceKeys.map(element => {
+    if (element.id === id) {
+      if (req.body.key) element.key = req.body.key
+      if (req.body.service) element.service = req.body.service
+      if (req.body.value) element.value = req.body.value
+      editedKey = element
+    }
```
songz: map should return something right? otherwise `updatedKeys` is undefined.

src/api/admin.ts
```diff
@@ -0,0 +1,80 @@
...
+app.put('/serviceHostsKey/:id', (req, res) => {
+  // replace servicekey info
+  let replacedKey
+  const serviceKeys = getData('serviceKeys')
+  const id: number = Number(req.params.id)
+  const updatedKeys = serviceKeys.map(element => {
+    if (element.id === id) {
+      element = { id, ...req.body }
+      replacedKey = element
+    }
```
songz: map is not returning anything.

src/api/admin.ts
```diff
@@ -0,0 +1,80 @@
...
+  })
+  setData('serviceKeys', updatedKeys)
+  res.json(deletedKey)
+})
+
+app.put('/serviceHostsKey/:id', (req, res) => {
+  // replace servicekey info
+  let replacedKey
+  const serviceKeys = getData('serviceKeys')
+  const id: number = Number(req.params.id)
```
songz: id should not be a number. Please refer to documentation: https://www.npmjs.com/package/uuid

src/api/admin.ts
```diff
@@ -0,0 +1,71 @@
...
+  const id: string = req.params.id
+  const updatedKeys = serviceKeys.map(element => {
+    if (element.id === id) {
+      if (req.body.key) element.key = req.body.key
+      if (req.body.service) element.service = req.body.service
+      if (req.body.value) element.value = req.body.value
+    }
+    return element
+  })
+  setData('serviceKeys', updatedKeys)
```
songz: Patch should respond with the updated element:
```
const updatedKey = servicekeys.find(e => e.id === id)
res.json(updatedKey)
```

src/api/admin.ts
```diff
@@ -0,0 +1,71 @@
...
+app.put('/serviceHostsKey/:id', (req, res) => {
+  // replace servicekey info
+  const serviceKeys = getData('serviceKeys')
+  const id: string = req.params.id
+  const updatedKeys = serviceKeys.map(element => {
+    if (element.id === id) element = { id, ...req.body }
+    return element
+  })
+  setData('serviceKeys', updatedKeys)
+  res.json(updatedKeys)
```
songz: PUT should respond with the updated element:
```
const updatedKey = servicekeys.find(e => e.id === id)
res.json(updatedKey)
```

src/api/admin.ts
```diff
@@ -0,0 +1,71 @@
...
+  )
+  res.json(selectedKey)
+})
+
+app.delete('/serviceHostsKey/:id', (req, res) => {
+  // delete a servicekey
+  const serviceKeys = getData('serviceKeys')
+  const updatedKeys = serviceKeys.filter(element => element.id !== req.params.id)
+  setData('serviceKeys', updatedKeys)
+  res.json(updatedKeys)
```
songz: PUT should respond with the deleted element, not the whole list:
```
const updatedKey = servicekeys.find(e => e.id === id)
res.json(updatedKey)
```

src/api/admin.ts
```diff
@@ -0,0 +1,71 @@
...
+  const serviceKeys = getData('serviceKeys')
+  res.json(serviceKeys)
+})
+
+app.get('/serviceHostsKey/:id', (req, res) => {
+  // grab one servicekey
+  const serviceKeys = getData('serviceKeys')
+  const selectedKey = serviceKeys.filter(
+    element => element.id === req.params.id
+  )
```
songz: filter returns an array, so this wouldn't work. You need `find`:
PUT should respond with the deleted element, not the whole list:
```
const selectedKey = servicekeys.find(e => e.id === id)
res.json(selectedKey)
```

PR #48, opened 2019-08-30
Fixing errors

allopez7:


Wolfy64:
Do not forget to `npm run autofix` before to push 🚀

src/api/admin.ts
```diff
@@ -12,11 +12,11 @@ const exec = util.promisify(cp.exec)
 app.post('/sslCerts', async (req, res) => {
   try {
     const serviceKeys = (getData('serviceKeys') || []).filter(
-      d => d.service === req.body.service
+      (d: any) => d.service === req.body.service
```
songz: should be `(d: ServiceKey)` no?

Also, could you change `getData` to `getProviders` or something? I think there was a helper function that returns the correct type.

src/providers/goDaddy.ts
```diff
@@ -5,8 +5,8 @@ import { ServiceKey } from '../api/types/admin'
 export const getDomains = () => {
   const serviceKeys: any = getData('serviceKeys')
```
songz: use helper function that returns the correct type

songz:
This is resolved by this PR: https://github.com/garageScript/myProxy/pull/56/files

PR #49, opened 2019-08-30
Adds POST to api/mappings

rkalra247:


src/api/index.ts
```diff
@@ -1,9 +1,11 @@
 import express from 'express'
 import adminRouter from './admin'
 import providers from './providers'
+import mappingRouter from './mapping'
 
 const apiRouter = express.Router()
 apiRouter.use('/admin', adminRouter.app)
 apiRouter.use('/providers', providers)
+apiRouter.use(mappingRouter)
```
songz: `apiRouter.use('/mappings', mappingRouter)`

src/api/mapping.ts
```diff
@@ -0,0 +1,17 @@
+import express from 'express'
+import uuid4 from 'uuid/v4'
+const mappingRouter = express.Router()
+const list =[]
+
+mappingRouter.post('/mappings', (req, res) =>{
```
songz: Look at comment above. If you are using `/mappings` up above, then this should be `post('/', (req, res)....`

src/api/mapping.ts
```diff
@@ -0,0 +1,17 @@
+import express from 'express'
+import uuid4 from 'uuid/v4'
+const mappingRouter = express.Router()
+const list =[]
```
songz: If you store in memory array, then everytime the app restarts, all data is lost. Please edit this file and add `mappings` to be stored: https://github.com/garageScript/myProxy/blob/master/src/api/lib/data.ts

src/api/mapping.ts
```diff
@@ -0,0 +1,17 @@
...
+
+mappingRouter.post('/mappings', (req, res) =>{
+  const domainObject = {
+    'domain': req.body.domain,
+    'port' : req.body.port,
+    'ip' : req.body.ip,
+    'id' : uuid4()
+  }
+  list.push(domainObject)
```
songz: don't use list, refer to this to see how `serviceKeys` is handling the `post` requests: https://github.com/garageScript/myProxy/blob/master/src/api/admin.ts#L37

src/api/lib/data.ts
```diff
@@ -2,7 +2,8 @@ import fs from 'fs'
 import { DB, ServiceKey } from '../types/admin'
 
 const data: DB = {
-  serviceKeys: []
+  serviceKeys: [],
+  domains: []
```
songz: you are storing the `mapping` resource, so `mappings` make more sense here.

songz: Could you create a helper function called `getMappings` here? so other developers can call the `getMappings` function.

src/api/mapping.ts
```diff
@@ -0,0 +1,20 @@
...
+const mappingRouter = express.Router()
+
+mappingRouter.post('/', (req, res) =>{
+  const domainKeys = getData('domains') || []
+  const domainObject = {
+    'domain': req.body.domain,
+    'port' : req.body.port,
+    'ip' : req.body.ip,
+    'id' : uuid4()
+  }
```
songz: could you create a `Mapping` type in `api/types.ts`

src/api/lib/data.ts
```diff
@@ -36,4 +37,8 @@ const getProviderKeys = (): Array<ServiceKey> => {
   return getData('serviceKeys')
 }
 
+const getMappings = (): Array<Mappings> => {
```
songz: Array of `Mapping` Objects. Singular type.

src/api/mapping.ts
```diff
@@ -0,0 +1,19 @@
+import express from 'express'
+import uuid4 from 'uuid/v4'
+import { setData, getData } from './lib/data'
+const mappingRouter = express.Router()
+
+mappingRouter.post('/', (req, res) => {
+  const domainKeys = getData('domains') || []
```
songz: why `domains`? Its `mappings`. Also, use your own helper function here: `getMappings()`

src/api/mapping.ts
```diff
@@ -0,0 +1,19 @@
...
+  const domainObject = {
+    domain: req.body.domain,
+    port: req.body.port,
+    ip: req.body.ip,
+    id: uuid4()
+  }
+  domainKeys.push(domainObject)
+  setData('domains', domainKeys)
```
songz: mappings

src/api/mapping.ts
```diff
@@ -0,0 +1,19 @@
+import express from 'express'
+import uuid4 from 'uuid/v4'
+import { setData, getData } from './lib/data'
+const mappingRouter = express.Router()
+
+mappingRouter.post('/', (req, res) => {
+  const domainKeys = getData('domains') || []
+  const domainObject = {
```
songz: `mappingObject: Mapping`

src/api/types/general.ts
```diff
@@ -0,0 +1,10 @@
+type Mappings = {
+  domain: string
+  port: string
+  ip: string
+  id: string
+}
+
+type DB = {
+  [mappings: string]: Array<Mappings>
+}
```
songz: why is `DB` type redefined here?

src/api/types/general.ts
```diff
@@ -0,0 +1,10 @@
+type Mappings = {
```
songz: Mapping

src/api/types/general.ts
songz: you should be exporting the type, so you can use it! look at this: https://github.com/garageScript/myProxy/blob/master/src/api/types/admin.ts

src/api/lib/data.ts
```diff
@@ -1,8 +1,10 @@
 import fs from 'fs'
 import { DB, ServiceKey } from '../types/admin'
+import { Mapping } from '../types/general'
 
 const data: DB = {
-  serviceKeys: []
+  serviceKeys: [],
+  mappings: []
```
songz: when the service restarts, your `mappings` value will always reset to `[]`. Then, you are not saving anything right? Look a few lines below, you will see `fs.readFile`. There, you need to do `data.mappings = fileData....`

src/api/mapping.ts
```diff
@@ -0,0 +1,21 @@
...
+import { setData, getData, getMappings } from './lib/data'
+import { Mapping } from './types/general' 
+const mappingRouter = express.Router()
+
+mappingRouter.post('/', (req, res) =>{
+  const domainKeys = getMappings()
+  const mappingObject: Mapping = {
+    'domain': req.body.domain,
+    'port' : req.body.port,
+    'ip' : req.body.ip,
```
songz: this should be defaulted to `127.0.0.1`, which is `localhost`. so... `'ip': req.body.ip || '127.0.0.1'`

PR #55, opened 2019-09-02
Admin UI: Adds functionality to create ServiceKeys

SahilKalra98:


src/public/serviceHostKeys.ts
```diff
@@ -53,9 +32,9 @@ class ServiceElement {
       return (
         acc +
         `
-          <div class="enteredKey">
-            ${key}
-            <input type="text" value=""></input>
+          <div class="enteredKeys">
+            <span id="serviceKeyName">${key}</span>
```
songz: don't use Ids. You can have many providers. each page should only have 1 element with id attribute, so don't use ids

src/public/serviceHostKeys.ts
```diff
@@ -64,9 +43,34 @@ class ServiceElement {
         <h4>${this.service.name}</h4>
         <li class="list-group-item">
           ${serviceKeys}
-          <button type="button" class="btn btn-primary">Create</button>
+          <button type="button" class="btn btn-primary" id="submitKeys">Create</button>
```
songz: don't use Ids. You can have many providers. each page should only have 1 element with id attribute, so don't use ids

src/public/serviceHostKeys.ts
```diff
@@ -53,9 +32,9 @@ class ServiceElement {
       return (
         acc +
         `
-          <div class="enteredKey">
-            ${key}
-            <input type="text" value=""></input>
+          <div class="enteredKeys">
+            <span class="serviceKeyName">${key}</span>
+            <input type="text" value="" class="keyInputs"></input>
```
songz: `keyInput`

src/public/serviceHostKeys.ts
```diff
@@ -53,9 +32,9 @@ class ServiceElement {
       return (
         acc +
         `
-          <div class="enteredKey">
-            ${key}
-            <input type="text" value=""></input>
+          <div class="enteredKeys">
```
songz: `providerContainer` is a better name IMO

src/public/serviceHostKeys.ts
```diff
@@ -64,9 +43,34 @@ class ServiceElement {
         <h4>${this.service.name}</h4>
         <li class="list-group-item">
           ${serviceKeys}
-          <button type="button" class="btn btn-primary">Create</button>
+          <button type="button" class="btn btn-primary submitKeys">Create</button>
```
songz: `createOrEditKeysButton`

PR #58, opened 2019-09-03
Moved provider into admin

Wolfy64:
Since now we show providers api keys in `api/providers`  I feel it's better to move this endpoint in `admin/providers`


songz:
there's a differenece.... `/api/admin/providers` should send back api data (JSON). `/admin/providers` should render providers UI. There's a difference.

PR #65, opened 2019-09-03
close #60 Save domain to fs after domain SSL certs has been created

### LINKED ISSUE #60: Save domain to fs after domain SSL certs has been created
Refer to this: https://github.com/garageScript/myProxy/pull/42/files
After an SSL certificate has been successfully generated for the selectedDomain, save the selectedDomain into fs under the key `availableDomains`. Each domain should have the properties: (with example)
```
[{
    domain: 'innout.life',
    expiration: <output from the ssl generation>,
    provider: 'dns_gd'
}]
```

allopez7:


src/api/admin.ts
```diff
@@ -26,6 +27,14 @@ app.post('/sslCerts', async (req, res) => {
     if (stderr) {
       console.log('stderr', stderr)
     }
+    const domains = getDomains()
```
songz: you don't want this to happen on error though. Right now you are saving domain even if there is error.

src/api/lib/data.ts
```diff
@@ -1,10 +1,11 @@
 import fs from 'fs'
 import { DB, ServiceKey } from '../types/admin'
-import { Mapping } from '../types/general'
+import { Mapping, Domain } from '../types/general'
 
 const data: DB = {
   serviceKeys: [],
-  mappings: []
+  mappings: [],
+  availableDomains: []
```
songz: when page loads, you need to set the initial value of this right?

songz: look at `readFile`

src/api/lib/data.ts
```diff
@@ -43,4 +44,8 @@ const getMappings = (): Array<Mapping> => {
   return getData('mappings')
 }
 
-export { getData, setData, getProviderKeys, getMappings }
+const getDomains = (): Array<Domain> =>{
```
songz: I would be consistent with naming and call this `getAvailableDomains`

src/api/admin.ts
```diff
@@ -24,10 +25,18 @@ app.post('/sslCerts', async (req, res) => {
       `${envVars} ./scripts/acme.sh/acme.sh --issue --dns ${req.body.service} -d ${req.body.selectedDomain} -d www.${req.body.selectedDomain}`
     )
     if (stderr) {
-      console.log('stderr', stderr)
+      return console.log('stderr', stderr)
```
songz: You are not sending back a response.
`return res.json({...})`

PR #66, opened 2019-09-04
Fix #64 UI for POST api/mappings

### LINKED ISSUE #64: UI creation page
As a developer, I would like to create a mapping for available domains.

Screen Shot 2019-09-03 at 1 42 51 PM

1. clicking on the button should submit a POST request `/mappings` to create a mapping
2. When mapping has been created, display the mappings in a list, like this image:
Screen Shot 2019-09-03 at 1 45 41 PM
3. (Separate Issue, should not be done as part of this issue). Dropdown must be populated by the Service in this issues:
https://github.com/garageScript/myProxy/issues/61
https://github.com/garageScript/myProxy/issues/60

rkalra247:


src/public/client.ts
```diff
@@ -0,0 +1,46 @@
+/* global */
+const getElement = (query: string, root?: HTMLElement): HTMLElement => {
```
songz: could you add a `//TODO: Move this function into separate helper function`

songz: its bad practice to copy / paste code from other files.

PR #91, opened 2019-09-06
Delete map fix #78

rkalra247:

PR #92, opened 2019-09-06
close #82 and fixes Bug when domains are not Authorized

### LINKED ISSUE #82: When new provider keys are created / updated, refresh the page
CUrrently, after updating or creating, there's 0 feedback so I have no idea what happened. (is it successful?)

SahilKalra98:


src/public/providers.ts
```diff
@@ -7,7 +7,7 @@ type Provider = {
   service: string
   keys: Array<ProviderKey>
   // eslint-disable-next-line
-  domains: Array<any>
+  domains: any
```
songz: domains should be an array right?

src/public/providers.ts
```diff
@@ -97,7 +102,9 @@ class ProviderElement {
         providerKeysContainer
       )
     })
-    provider.domains.map(domain => {
+    const isAuthorized = !provider.domains.code ? true : false
```
songz: what is this doing?

src/public/providers.ts
```diff
@@ -97,7 +102,9 @@ class ProviderElement {
...
+    const providerDomains = isAuthorized ? provider.domains : []
```
songz: why do we need this? `/api/admin/providers` should always be be returning an array of domains. If its not, then the fix must go in at the api endpoint to return `[]` if domains does not exist.

src/public/providers.ts
```diff
@@ -64,7 +69,7 @@ class ProviderKeyElement {
         headers: {
           'Content-Type': 'application/json'
         }
-      })
+      }).then(() => window.location.reload())
```
songz: instead of reloading the page, should we have a `render` function so that we are consistent with implementation across different pages? For example, the `render` function in this PR: https://github.com/garageScript/myProxy/pull/91/files

src/api/admin.ts
```diff
@@ -123,7 +123,12 @@ app.patch('/providerKeys/:id', (req, res) => {
 
 app.get('/providers', async (_, res) => {
   const data = await Promise.all([goDaddy.getDomains()])
-  return res.json(data) // Data send to view providers ?
+  console.log(data)
```
songz: remove

Issue #90, opened 2019-09-06
Proxy


songz:
Grouped by PR:

1. `app.listen` should only be called if `ADMIN` environment is present. If not present, `app.listen` should not happen, and it should log message: 'Admin UI/API is turned off'
2. Create a HTTPS server, listening to port 443. Handle SNICallback to send the ssl certificates. For all requests, just send back "Hello World"
3. From the hostname, lookup `mappings` db to find the correct port, then proxy to the port.
4. Create an `express` server to listen to http requests on port 80. Redirect every request to https domain (make sure path and query parameters are included in the redirection).

songz:
Resolved: https://github.com/garageScript/myProxy/pull/183/files

PR #94, opened 2019-09-07
#90 prevent users from accessing admin without input of ADMIN password

allopez7:


src/app.ts
```diff
@@ -28,4 +28,11 @@ app.post('/login', (req, res) => {
   res.redirect('/admin')
 })
 
-app.listen(port, () => console.log(`app listening on port ${port}!`))
+const listener = (): void =>{
+  if(!process.env.ADMIN){
+    return console.log('Admin UI/API is turned off')
+  }
+  app.listen(port, () => console.log(`app listening on port ${port}!`))
```
songz: could you also change `package.json` so that `yarn start` will set ENV variables? This way, ppl won't be confused when this gets merged in.

Issue #27, opened 2019-08-20
Overview


songz:
# Goal

## Part 1: Admin
1. ~~Admin: Users should be able to save their serviceKeys so that myProxy can use it to retrieve their domains and verify ssl challenge.~~
> ~~User logs in to their GoDaddy account, finds their apiKey/secret. Copies them and paste it so that we can save it.~~
> ~~Example:~~
```
      service: GoDaddy.   
      key1: API_KEY.   
      key2: API_SECRET.   
```
2. ~~After user saves the service Keys, we hit the service to retrieve all the domains that they own. Next to each domain, there will be a button `setup`~~
3. ~~When user clicks on the `setup` button, we do 2 things:~~
> 1. ~~run acme.sh script with service keys.~~
> 2. ~~Update the `CNAME` records for the domain using the service APIs so that all requests going to the domain will go to the server running myproxy~~

## Part 2: User
1. ~~User goes to application page. All domains that are `setup` will be in the dropdown. User can type in any subdomains~~
2. ~~When user creates an domain, create a port for them automatically that they can run their apps on.~~

## Part 3: Proxy 
1. ~~When request comes in to https domain, find the port and forward it accordingly.~~

songz:
DONE! - 9/23

PR #96, opened 2019-09-07
Overview Part 1 .3

Wolfy64:
**Overview** Part 1.3 #27 
Create a **ssl certificate** works but I exceeded the limit of [5 per week](https://letsencrypt.org/docs/rate-limits/) 😇
Setup a new **record** (name, ip) with goddady API work.


src/api/admin.ts
```diff
@@ -122,7 +127,7 @@ app.patch('/providerKeys/:id', (req, res) => {
 })
 
 app.get('/providers', async (_, res) => {
-  const data = await Promise.all([goDaddy.getDomains()])
+  const data = await Promise.all([providers.dns_gd.getDomains()])
```
Wolfy64: I got a TS warning `Property 'getDomains' does not exist on type 'Function'.ts(2339)`
I don't know how to solve it

src/api/admin.ts
```diff
@@ -26,6 +27,12 @@ app.post('/sslCerts', async (req, res) => {
 
     if (stderr) return res.json({ stderr })
 
+    const { stdout: ipaddress } = await exec('curl ifconfig.me')
+    const setRecords: Function = await providers[service].setRecord(
```
Wolfy64: I got a TS warning `Element implicitly has an 'any' type because expression of type 'any' can't be used to index type '{ dns_gd: Function; }'.ts(7053)`
I don't know how to solve it

src/providers/goDaddy.ts
```diff
@@ -1,56 +1,44 @@
...
+    const serviceKeys = getProviderKeys()
+    return (
+      serviceKeys.find(k => k.service === 'dns_gd' && k.key === keyInfo.key) ||
+      keysDefault
+    )
+  })
+  return keys as ServiceKey[]
+}
+const findKey = (key: string): string => {
+  return getKeys().find(k => k.key === key).value
```
Wolfy64: I got a TS warning `Object is possibly 'undefined'.ts(2532)`
I don't get it because I typed everything

src/providers/goDaddy.ts
```diff
@@ -1,56 +1,44 @@
...
 import { getProviderKeys } from '../api/lib/data'
 import { Provider } from '../api/types/general'
+import { ServiceKey } from '../api/types/admin'
 
-export const getDomains = async (): Promise<Provider> => {
-  const service = 'https://api.godaddy.com'
-  const name = 'Godaddy'
-  const serviceKeys = getProviderKeys()
+const name = 'Godaddy'
+const service = 'https://api.godaddy.com'
```
Wolfy64: I wonder if this kind on `constant` would be a best practice in uppercase or not?

src/api/admin.ts
songz: `setRecords` is not a function. it is data returned from the API call

src/providers/goDaddy.ts
songz: yeah, uppercase is best.

src/api/admin.ts
```diff
@@ -1,30 +1,56 @@
...
+    message: 'SSL Certs and domain name records successfully created'
+  }
   try {
     const serviceKeys = getProviderKeys().filter(d => d.service === service)
     const { keys } = serviceConfig[service]
     const envVars = keys.reduce((acc: string, key: string) => {
       const { value } = serviceKeys.find(d => d.key === key) || { value: '' }
       return acc + `${key}=${value} `
     }, '')
-    const { stdout, stderr } = await exec(
```
allopez7: If stdout is deleted here. How will expiration date, key and cert be obtained? Is line 40-50 the setup for information storage.

PR #101, opened 2019-09-08
fix #100 UI for edit/PATCH

### LINKED ISSUE #100: Fixing UI for Edit/Patch
https://cl.ly/e92a9c872b42

rkalra247:
https://cl.ly/6f9903f8712f

### DIFF
--- a/src/public/client.ts
+++ b/src/public/client.ts
@@ -12,7 +12,6 @@ const domainList: HTMLElement = helper.getElement('.domainList')
 const dropDownDomains: HTMLElement = helper.getElement('.dropdown-menu')
 let selectedHost = ''
 
-// eslint-disable-next-line
 class DomainMap {
   constructor(data: Mapping) {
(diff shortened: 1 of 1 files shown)


src/public/client.ts
```diff
@@ -63,20 +62,29 @@ class DisplayMap {
...
-         Domain: <input class=' form-control domain'type="" value='${data.domain}'>
-         Port:  <input class='form-control port'type="" value=${data.port}>
-         IP:  <input class='form-control ip' type="" value=${data.ip}>
-         <hr/>
-         <div class ='saveButtonContainer'>
-         <button  class='btn save' style="padding: 0px 0px 0px 15px">SAVE</button>
-         </div>
-         </li>
+        <li class='list-group-item'>
+        <form>
```
Wolfy64: I wonder if it would not be better to use onsubmit instead of onclick with a `<button type="submit">` ?

PR #104, opened 2019-09-09
Bug: SSL Cert needs to send two commands for *.domain and .domain

### LINKED ISSUE #102: Bug: SSL Cert needs to send two commands for  *.domain and .domain

Wolfy64:
Fix #102 
After reading the doc, if I don't misunderstand the first parameter has to be the **Main_domain** other parameters are **SAN_Domains**.
I can't test it due to the limitation as 5 by week however I can run `acme.sh --list` that show the list

### DIFF
--- a/src/api/admin.ts
+++ b/src/api/admin.ts
@@ -25,10 +25,10 @@ app.post('/sslCerts', async (req, res) => {
       const { value } = serviceKeys.find(d => d.key === key) || { value: '' }
       return acc + `${key}=${value} `
     }, '')
-    const { stderr } = await exec(
-      `${envVars} ./acme.sh/acme.sh --issue --dns ${service} -d ${selectedDomain} -d "*.${selectedDomain}" --force`
-    )
-
+    const acme = `./acme.sh/acme.sh --issue --dns ${service}`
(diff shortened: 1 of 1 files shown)


src/api/admin.ts
```diff
@@ -26,7 +26,7 @@ app.post('/sslCerts', async (req, res) => {
       return acc + `${key}=${value} `
     }, '')
     const { stderr } = await exec(
-      `${envVars} ./acme.sh/acme.sh --issue --dns ${service} -d "*.${selectedDomain}" -d ${selectedDomain} --force`
+      `${envVars} ./acme.sh/acme.sh --issue --dns ${service} -d ${selectedDomain} -d "*.${selectedDomain}" --force`
```
songz: when you type this into the command line, it only creates one cert. Here's what I think we need to do:
```
const { stderr } = await exec(
  `${envVars} ./acme.sh/acme.sh --issue --dns ${service} -d ${selectedDomain} --force`
)
... await exec(
  `${envVars} ./acme.sh/acme.sh --issue --dns ${service} -d *.${selectedDomain} --force`
)
```

PR #105, opened 2019-09-10
Adds Git Hooks functionality

SahilKalra98:


src/api/mapping.ts
```diff
@@ -15,6 +19,21 @@ mappingRouter.post('/', (req, res) => {
   }
   domainKeys.push(mappingObject)
   setData('mappings', domainKeys)
+
+  // WIP: Move to a script folder
+  const projectFolder = `${path.join(__dirname, '../../projects')}/${
+    req.body.domain
```
songz: `req.body.domain + req.body.subdomain`

src/api/mapping.ts
```diff
@@ -15,6 +19,21 @@ mappingRouter.post('/', (req, res) => {
...
+  }`
+  exec(`
+    mkdir ${projectFolder}
+    git init ${projectFolder}
+    cd ${projectFolder} 
+    cp ../../scripts/post-receive .git/hooks/
+    chmod 755 .git/hooks/post-receive
```
songz: `chmod +x` doesn't work? must we use `755`?

src/api/mapping.ts
```diff
@@ -15,6 +19,21 @@ mappingRouter.post('/', (req, res) => {
   }
   domainKeys.push(mappingObject)
   setData('mappings', domainKeys)
+
+  // WIP: Move to a script folder
+  const projectFolder = `${path.join(__dirname, '../../projects')}/${
```
songz: why do you need `../../`? Why not put it in this folder? Like `./projects`

src/api/mapping.ts
```diff
@@ -15,6 +19,18 @@ mappingRouter.post('/', (req, res) => {
   }
   domainKeys.push(mappingObject)
   setData('mappings', domainKeys)
+  const projectFolder = `${path.join(__dirname, '../../projects')}/${req.body
+    .domain + req.body.subDomain}`
+  exec(`
+    mkdir -p ../../projects
+    mkdir ${projectFolder}
```
songz: `mkdir ${req.body.domain}.${req.body.subDomain`

or

`mkdir ${req.body.domain}\.${req.body.subDomain`

Please find out which one works better.

scripts/post-receive
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+cd ..
+GIT_DIR='.git'
+umask 002 && git reset --hard
+git checkout master
+git branch -D prod
+git checkout -b prod
+#pm2 startOrRestart ../../deploy.config.js
```
songz: y is this commented out? it should be there

scripts/post-receive
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+cd ..
```
songz: why `cd ..`?

src/api/mapping.ts
```diff
@@ -15,7 +20,40 @@ mappingRouter.post('/', (req, res) => {
   }
   domainKeys.push(mappingObject)
   setData('mappings', domainKeys)
-  res.json(mappingObject)
+  const projectFolder = `${req.body.domain}`
```
songz: You also need to add in the subdomain. You can create N number of repos per domain. not just 1 domain.

src/api/mapping.ts
```diff
@@ -15,7 +20,40 @@ mappingRouter.post('/', (req, res) => {
...
+  const prodConfig = {
+    apps: [
+      {
+        name: 'myProxy-prod',
```
songz: name of the app should be their projectName (subdomain + domain)

src/api/mapping.ts
```diff
@@ -15,7 +20,40 @@ mappingRouter.post('/', (req, res) => {
...
+        script: './build/app.js',
+        instances: 1,
+        autorestart: true,
+        watch: false,
+        max_memory_restart: '1G',
+        env_production: {
+          NODE_ENV: 'production',
+          PORT: parseInt(req.body.port, 10),
+          ADMIN: '123'
```
songz: remove admin, they should set this up themselves right? or use `process.env.ADMIN`

src/api/mapping.ts
```diff
@@ -15,7 +20,40 @@ mappingRouter.post('/', (req, res) => {
...
+        }
+      }
+    ]
+  }
+  exec(`
+    mkdir -p ~/projects
```
songz: `const projectPath = ....` and then use `projectPath` instead of hardcoding the `~/projects` everywhere.

src/api/mapping.ts
```diff
@@ -15,7 +20,40 @@ mappingRouter.post('/', (req, res) => {
...
+    mkdir ~/projects/${projectFolder}
+    git init ~/projects/${projectFolder}
+    cp ${path.join(
+    __dirname,
+    '../../scripts'
+  )}/post-receive ~/projects/${projectFolder}/.git/hooks/
+    cd ~/projects/${projectFolder}
+    echo "module.exports =" > deploy.config.js
+    echo '${JSON.stringify(prodConfig)}' >> deploy.config.js
```
songz: could you combine this into 1 command? 
`echo 'module.exports = JSON.stringify(prodConfig)'`

src/api/mapping.ts
```diff
@@ -15,7 +20,39 @@ mappingRouter.post('/', (req, res) => {
   }
   domainKeys.push(mappingObject)
   setData('mappings', domainKeys)
-  res.json(mappingObject)
+  const projectFolder = `${req.body.domain + req.body.subDomain}`
```
```
`${req.body.subDomain}.${req.body.domain}`
```

songz: for example: `sahil.mcgiggle.com`

songz: `sahil` is subdomain, `mcgiggle.com` is domain.

songz: a better name for this is `fullDomain`. Its technically not a folder.

src/api/mapping.ts
```diff
@@ -15,7 +20,39 @@ mappingRouter.post('/', (req, res) => {
...
+  }
+  const projectPath = '~/projects'
+  exec(`
+    mkdir -p ${projectPath}
+    mkdir ${projectPath}/${projectFolder}
+    git init ${projectPath}/${projectFolder}
+    cp ${path.join(
+    __dirname,
+    '../../scripts'
+  )}/post-receive ${projectPath}/${projectFolder}/.git/hooks/
```
songz: this is a little confusing, could you create a variable? `const postReceiveScriptPath = ....` and then use the variable here.

src/api/mapping.ts
```diff
@@ -15,7 +20,37 @@ mappingRouter.post('/', (req, res) => {
   }
   domainKeys.push(mappingObject)
   setData('mappings', domainKeys)
-  res.json(mappingObject)
+  const fullDomain = `${req.body.subDomain}.${req.body.domain}`
+  const prodConfig = {
+    apps: [
+      {
+        name: `${fullDomain}`,
```
songz: `name: fullDomain`, because its already a string.

src/api/mapping.ts
```diff
@@ -15,7 +20,37 @@ mappingRouter.post('/', (req, res) => {
...
+        max_memory_restart: '1G',
+        env_production: {
+          NODE_ENV: 'production',
+          PORT: parseInt(req.body.port, 10)
+        }
+      }
+    ]
+  }
+  const projectPath = '~/projects'
+  const postReceiveScriptPath = `${path.join(__dirname, '../../scripts')}`
```
songz: `path.join(....)` is already a string right? Then its pointless to add `${....}`

PR #108, opened 2019-09-11
https server for securecontext 

allopez7:
…nd logname into readdir

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,6 @@
     "eslint": "^6.2.2",
     "eslint-config-prettier": "^6.1.0",
     "jest": "^24.8.0",
-    "nodemon": "^1.19.2",
     "nyc": "^14.1.1",
     "prettier": "^1.18.2",
     "source-map-support": "^0.5.13",
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "express": "^4.17.1",
     "husky": "^3.0.3",
     "node-fetch": "^2.6.0",
-    "nodemon": "^1.19.2",
     "pretty-quick": "^1.11.1",
     "uuid": "^3.3.3"
   },
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "express": "^4.17.1",
     "husky": "^3.0.3",
     "node-fetch": "^2.6.0",
+    "nodemon": "^1.19.2",
     "pretty-quick": "^1.11.1",
     "uuid": "^3.3.3"
   },
--- a/src/app.ts
+++ b/src/app.ts
@@ -45,10 +45,14 @@ app.post('/login', (req, res) => {
 })
 
 const listener = (): void => {
+  const cyan = '\x1b[36m\u001b[1m%s\x1b[0m'
+  const red = '\x1b[31m\u001b[1m%s\x1b[0m'
   if (!process.env.ADMIN) {
-    return console.log('Admin UI/API is turned off')
+    return console.log('red', 'Admin UI/API is turned off')
   }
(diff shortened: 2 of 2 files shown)


src/app.ts
```diff
@@ -35,4 +38,15 @@ const listener = (): void => {
   app.listen(port, () => console.log(`app listening on port ${port}!`))
 }
 
+const server = https.createServer({
+  SNICallback: (domain, cb) => {
+    const adminUser = execSync('logname').toString()
+    fs.readdir(`/home/albertoe/\.acme\.sh`, (err, files)=>{
```
allopez7: this is causing a warning currently, will update soon.

package.json
```diff
@@ -36,6 +36,7 @@
     "express": "^4.17.1",
     "husky": "^3.0.3",
     "node-fetch": "^2.6.0",
+    "nodemon": "^1.19.2",
```
songz: remove.

src/app.ts
```diff
@@ -35,4 +38,15 @@ const listener = (): void => {
   app.listen(port, () => console.log(`app listening on port ${port}!`))
 }
 
+const server = https.createServer({
+  SNICallback: (domain, cb) => {
+    const adminUser = execSync('logname').toString()
```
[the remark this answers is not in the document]
allopez7: gives me 'root'

package.json
allopez7: was throwing error unless I added.

songz: this should be a dev dependency. Could you remove, then run `yarn add nodemon -D`. In production, we will not be using nodemon

songz: since you fixed this in a different PR, could you remove this? https://github.com/garageScript/myProxy/pull/123

allopez7: ok

src/app.ts
```diff
@@ -41,14 +45,38 @@ app.post('/login', (req, res) => {
 })
 
 const listener = (): void => {
-  const cyan = '\x1b[36m\u001b[1m%s\x1b[0m'
```
allopez7: I did not delete these, pulled from main branch and this is the result.

src/app.ts
```diff
@@ -41,14 +45,38 @@ app.post('/login', (req, res) => {
...
-  const red = '\x1b[31m\u001b[1m%s\x1b[0m'
```
songz: why did you remove all these?

allopez7: I did not remove them. Updated my branch and this happened.

src/app.ts
```diff
@@ -51,4 +55,32 @@ const listener = (): void => {
...
+    SNICallback: (domain, cb) => {
+      // using whoami will return root because app is run with sudo
+      const adminUser = execSync('logname')
+        .toString()
+        .trim()
+      // escape characters required or readFileSync will not find file
+      const secureContext = tls.createSecureContext({
+        /* eslint-disable */
+        key: fs.readFileSync(
+          `/home/${adminUser}/\.acme\.sh/*\.${domain}/*\.${domain}\.key`
```
songz: instead of `/home/${adminUser}`, could you do `~/` instead? Because for `root` user, the home path is `/root`

src/app.ts
```diff
@@ -5,6 +5,10 @@ import crypto from 'crypto'
 import cookieParser from 'cookie-parser'
 import { adminRouter } from './admin/index'
 import { apiRouter } from './api/index'
+import https from 'https'
+import fs from 'fs'
+import { execSync } from 'child_process'
```
songz: `execSync` is not used here.

Issue #112, opened 2019-09-12
Throw a error on the first app launch


Wolfy64:
Throw a error on the first app launch on `npm run start` just after `npm run setup`

```js
internal/modules/cjs/loader.js:716
    throw err;
    ^

Error: Cannot find module '/home/dev/david/app.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:713:15)
    at Function.Module._load (internal/modules/cjs/loader.js:618:27)
    at Function.Module.runMain (internal/modules/cjs/loader.js:931:10)
    at internal/main/run_main_module.js:17:11 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```

PR #113, opened 2019-09-12
Fix bug #112

Wolfy64:
Fix bug #112


scripts/setup.sh
```diff
@@ -5,4 +5,5 @@ if [ ! -d "./acme.sh" ] ; then
   git clone https://github.com/Neilpang/acme.sh.git
 fi
 cd ./acme.sh
-./acme.sh --install
\ No newline at end of file
+./acme.sh --install
+npm run build
```
songz: just to be safe, should you `cd ../` before `npm run build`?

PR #118, opened 2019-09-13
Add auth middleware

### LINKED ISSUE #89: BUG: api requests times out if users are not admin
To reproduce:
Delete all cookies. 
Screen Shot 2019-09-06 at 10 23 51 AM

Go to root path.
Look at network tab
API request to `/api/mappings` does not receive response.

---
Expected behavior:
Should return status code 401 for protected endpoints (`/api/admin`)
Should not fail for non protected endpoints. (i.e.`/api/mappings`)

Wolfy64:
Fixed #89
Add auth middleware and hash password

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,6 +1,7 @@
 import 'dotenv/config'
 import express from 'express'
 import path from 'path'
+import crypto from 'crypto'
 import cookieParser from 'cookie-parser'
 import { adminRouter } from './admin/index'
 import { apiRouter } from './api/index'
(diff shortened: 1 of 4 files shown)


src/app.ts
```diff
@@ -22,10 +23,21 @@ app.get('/', (req, res) => res.render('client'))
 app.get('/login', (req, res) => res.render('login', { error: '' }))
 
 app.post('/login', (req, res) => {
-  if (process.env.ADMIN !== req.body.adminPass)
-    return res.render('login', { error: 'Wrong Admin Password' })
-  res.cookie('adminPass', req.body.adminPass)
-  res.redirect('/admin')
+  const hashPass = (password: string): string => {
```
songz: these should be moved into the `auth` folder, created a separate refactoring issue: https://github.com/garageScript/myProxy/issues/120

PR #119, opened 2019-09-13
Comment unused code

Wolfy64:
Comment unused code to remove **Typescript warning** until that feature is done.


songz:
we need this, perhaps you could send a fetch request to `/api/availableDomains` to populate the dropdown.

Issue #129, opened 2019-09-14
Re-organize API folder


Wolfy64:
> api/ folder is confusing... inside we have all routes (except /admin/providers), lib/ types/ folder...
We might add a routes folder with api.ts and admin.ts and move out lib/,types into /src ?

> lib and types should be in src folder

PR #130, opened 2019-09-14
⚠️ Refactor Move lib/ into ./src

Wolfy64:
WIP #129 
> lib and types should be in src folder

I will refactor step by step to avoid breaking change 💥


songz:
this will have even more merge conflicts as we resolve all the existing typescript errors, so closing this for now. Rework / reopen after typescript errors are resolved.

PR #131, opened 2019-09-14
fix #103 add subDomain to POST/api/mappings

### LINKED ISSUE #103: Fix name for subdomain, remove domain
It is currently storing subdomain value. Domain value will be loaded from different api.

rkalra247:


src/public/client.ts
```diff
@@ -168,11 +168,13 @@ create.onclick = (): void => {
   const portValue = port.value
   const domain = subDomain.value + selectedHost
```
songz: domain is just `selectedHost`

PR #138, opened 2019-09-15
⚠️ Issue 129 Please do not merge before test it

### LINKED ISSUE #129: Re-organize API folder

Wolfy64:
# Please test it before to accept
close #129 Refactor `api/` folder
Move folders `lib/` and `types/` into `./src`

I run it and test it manually **everything work** without or breaking change in my local.


src/public/client.ts
```diff
@@ -1,11 +1,5 @@
 /* global helper */
-type Mapping = {
-  domain: string
-  subDomain: string
-  ip: string
-  port: string
-  id: string
-}
+import { Mapping } from '../types/client'
```
songz: I don't think import works for client side builds yet. Does this work?

songz:
verified, not causing issues.

PR #139, opened 2019-09-15
Add script command `npm run server`

### LINKED ISSUE #128: Question about prod.config.js
Does it make sense for a user that not use `pm2` to have this file? 
I think this file should not be on Github only on the server same for `deploy.sh`. 
**OR**
We can add `pm2` as dependency and change `npm start` to run it ?

```
module.exports = {
  apps: [
    {
      name: 'myProxy-prod',
      script: './build/app.js',
      instances: 1,
      autorestart: true,
      watch: false,
      max_memory_restart: '1G',
      env_production: {
        NODE_ENV: 'production', 
        PORT: 3000, // Should be setup with Process.env.PORT
        ADMIN: '123' // Should be setup with Process.env.ADMIN
      },
    },
  ],
}
```

Wolfy64:
`npm run server` will run pm2
`npm run server:stop` stop server

close #128


.eslintrc.json
```diff
@@ -1,6 +1,6 @@
 {
     "env": {
-        "browser": true,
+        "node": true,
```
Wolfy64: Fixed ESLint warning `'process' is not defined` for `process.env` within  `scripts/prod.config.js`
Thanks to this [StackOverflow](https://stackoverflow.com/questions/50894000/eslint-process-is-not-defined) thread

src/app.ts
```diff
@@ -51,30 +51,30 @@ const listener = (): void => {
   )
 }
```
Wolfy64: I didn't touch this code below. Running `autofix` format the code...

package.json
```diff
@@ -13,7 +13,9 @@
     "prettier:write": "npm run prettier:base -- --write \"src/**/*.{ts,tsx}\"",
     "coverage": "nyc npm run test",
     "eslint:fix": "./node_modules/.bin/eslint src/ --ext .ts --fix",
-    "autofix": "npm run prettier:write; npm run eslint:fix"
+    "autofix": "npm run prettier:write; npm run eslint:fix",
+    "server": "pm2 startOrRestart ./scripts/prod.config.js --env production --update-env",
```
songz: this should run setup and everything.... ie:
```
npm run setup
npm install
npm run build
pm2 startOrRestart...
```

package.json
```diff
@@ -36,6 +38,7 @@
     "express": "^4.17.1",
     "husky": "^3.0.3",
     "node-fetch": "^2.6.0",
+    "pm2": "^3.5.1",
```
songz: should not be in node modules. `pm2` should be install globally during setup step

scripts/deploy.sh
```diff
@@ -3,4 +3,4 @@
 cd /home/dev/prodProxy
 git pull origin master
 npm run setup
```
songz: `npm run setup` could be part of the `server` instructions, so this can also be removed.

package.json
Wolfy64: Why we can't install it in `node_modules` ? 
To install `pm2` globally we have to be log as root or ask the final user to run sudo... (not good)
Or we have to tell them within Readme instruction.

scripts/setup.sh
```diff
@@ -1,11 +1,15 @@
 #!/bin/bash
 
 npm install
+npm install pm2 -g
```
Wolfy64: See comment below for sudo or not ⬇️

PR #141, opened 2019-09-15
close #132 SSl Cert Generation

### LINKED ISSUE #132: BUG: SSL Cert generation is not working
Please fix this API endpoint. Its a blocker, because without `availableDomains` nothing works. https://github.com/garageScript/myProxy/blob/master/src/api/admin.ts#L14

Update 9/16
POST `/sslCerts` is still not creating certificates in production. Please take note of the picture below:
1. When you test, please remove the existing `key`/ `cert`. SSL cert generation takes > 30 seconds per cert. So if we are generating 2 certs (root and wildcard), the post request should take 60 seconds.
2. We are chaining the cert generation command using `&`. This means background, so the env variables set is not applied. I would use `;` instead. 
3. Make sure the post request does not time out (possible up to 1 minute wait).

Screen Shot 2019-09-16 at 1 54 38 PM

SahilKalra98:


src/helpers/httpRequest.ts
```diff
@@ -1,12 +1,8 @@
 import fetch from 'node-fetch'
 
 const sendRequest = async <T>(url: string, options: object): Promise<T> => {
-  try {
-    const request = await fetch(url, options)
-    return await request.json()
-  } catch (error) {
-    throw new Error(error)
-  }
```
SahilKalra98: https://github.com/github/fetch/issues/268, response.json() could sometimes have parsing errors

PR #142, opened 2019-09-16
Fix getProviderKeys and getMappings bug

Wolfy64:
`getProviderKeys` & `getMappings` was fetching the wrong table

PR #144, opened 2019-09-16
Fetch Dropdown

Wolfy64:
Fetch Dropdown into `/admin` This PR rely on this [PR](https://github.com/garageScript/myProxy/pull/142)


src/api/index.ts
```diff
@@ -1,11 +1,17 @@
 import express from 'express'
 import adminRouter from './admin'
 import mappingRouter from './mapping'
+import { getAvailableDomains } from './lib/data'
```
allopez7: I think this is from the last PR.

Wolfy64: @allopez7  Correct I should precise that PR rely on #143

PR #161, opened 2019-09-17
CNAME record not setup correctly

### LINKED ISSUE #157: CNAME record not setup correctly
CNAME should be `*`, not the domain name.

Example:
Screen Shot 2019-09-16 at 8 45 44 PM

Wolfy64:
FIxed #157 
I only worked on `src/providers/goDaddy.ts` other change are from `npm run autofix` 
To make plain-fully work `api/admin/sslCerts` we have to fix this bug first #162

### DIFF
--- a/src/auth/index.ts
+++ b/src/auth/index.ts
@@ -10,11 +10,7 @@ const isCorrectCredentials = (password: string): boolean => {
 
 auth.use('/*', (req, res, next) => {
   const { adminPass } = req.cookies
-  if (
-    !adminPass &&
-    !isCorrectCredentials((req.headers.Authorization as string) || '')
-  )
-    return res.status(401).redirect('/login')
+  if (!adminPass && !isCorrectCredentials(req.headers.Authorization as string || '')) return res.status(401).redirect('/login')
   next()
 })
 
--- a/src/auth/index.ts
+++ b/src/auth/index.ts
@@ -10,7 +10,11 @@ const isCorrectCredentials = (password: string): boolean => {
 
 auth.use('/*', (req, res, next) => {
   const { adminPass } = req.cookies
-  if (!adminPass && !isCorrectCredentials(req.headers.Authorization as string || '')) return res.status(401).redirect('/login')
+  if (
+    !adminPass &&
+    !isCorrectCredentials((req.headers.Authorization as string) || '')
+  )
+    return res.status(401).redirect('/login')
   next()
 })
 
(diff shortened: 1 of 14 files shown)


src/providers/goDaddy.ts
```diff
@@ -58,14 +58,24 @@ export const setRecord = async (
...
+    },
+    {
+      data: 'ns01.domaincontrol.com',
+      name: '',
+      type: 'NS'
+    },
+    {
+      data: 'ns02.domaincontrol.com',
+      name: '',
+      type: 'NS'
```
songz: should not be replacing nameservers

src/providers/goDaddy.ts
```diff
@@ -58,14 +58,24 @@ export const setRecord = async (
...
-      data: domain,
-      name: domain,
+      data: '@',
+      name: '*',
       type: 'CNAME'
+    },
+    {
+      data: 'ns01.domaincontrol.com',
+      name: '',
+      type: 'NS'
```
songz: should not be replacing nameservers

src/providers/goDaddy.ts
```diff
@@ -58,14 +58,24 @@ export const setRecord = async (
...
+    {
+      data: 'ns02.domaincontrol.com',
+      name: '',
+      type: 'NS'
     }
   ]
 
   const options = {
-    method: 'PATCH',
+    method: 'PUT',
```
songz: I don't think `PUT` is a good idea, this would get rid of existing `MX` records and stuff.
Steps:
1. Get existing A and CNAME Records.
2. Patch / Delete problematic records.
`A` record should have only one `@` entry.
`CNAME` record must have only one `*` entry, pointing to `@`.

src/auth/index.ts
```diff
@@ -10,7 +10,11 @@ const isCorrectCredentials = (password: string): boolean => {
 
 auth.use('/*', (req, res, next) => {
   const { adminPass } = req.cookies
-  if (!adminPass && !isCorrectCredentials(req.headers.Authorization as string || '')) return res.status(401).redirect('/login')
```
Wolfy64: From npm run autofix...

src/types/general.ts
```diff
@@ -34,11 +34,19 @@ type ServiceConfig = {
   dns_gd: ProviderService
 }
 
+type NameServer = {
```
songz: this is godaddy specific, should not be in general types.

songz: because if every domain provider created their own types, it'll get messy.

Wolfy64: Sure make sense

songz:
closing, please refer to this: https://github.com/garageScript/myProxy/pull/174

PR #166, opened 2019-09-17
fix #165 setUp default value for IP to 127.0.0.1

### LINKED ISSUE #165: Set up the default IP

rkalra247:


src/api/mapping.ts
```diff
@@ -11,6 +11,7 @@ const exec = util.promisify(cp.exec)
 
 mappingRouter.post('/', (req, res) => {
   const domainKeys = getMappings()
+  if(!req.body.ip) req.body.ip = '127.0.0.1'
```
songz: `req.body` is part of the request, and request does not have `req.body.ip`, so you should keep it that way. 

Instead, below you should just do `ip: req.body.ip || '127.0.0.1'`

PR #168, opened 2019-09-18
fix #109 Automate Ports if no value is assigned to port

### LINKED ISSUE #109: Automate Ports POST/api/mappings

rkalra247:


src/api/mapping.ts
```diff
@@ -8,13 +8,18 @@ import { setData, getMappings } from '../lib/data'
 import { Mapping } from '../types/general'
 const mappingRouter = express.Router()
 const exec = util.promisify(cp.exec)
+let portCounter = 3001
 
 mappingRouter.post('/', (req, res) => {
   const domainKeys = getMappings()
+  if (req.body.port === '') {
+    portCounter += 1
```
songz: hmmm.... everytime our server restarts, it will be 3002.

src/api/mapping.ts
```diff
@@ -8,13 +8,19 @@ import { setData, getMappings } from '../lib/data'
 import { Mapping } from '../types/general'
 const mappingRouter = express.Router()
 const exec = util.promisify(cp.exec)
+let portCounter = 3002
 
 mappingRouter.post('/', (req, res) => {
   const domainKeys = getMappings()
+  if(req.body.port === ''){
+    domainKeys.forEach((e) => {
+      if(e.port === portCounter.toString()) portCounter += 1
```
SahilKalra98: Maybe sort the keys by port number, and move portCounter into mappingRouter.post?

SahilKalra98: if domainKeys was [{port: 3003 ....}, {port: 3002 ...}, {port: 3008 ...}] then when it goes through the forEach it will create another domain at port: 3003

PR #175, opened 2019-09-21
Fixes CNAME records

SahilKalra98:


src/providers/goDaddy.ts
```diff
@@ -2,6 +2,7 @@ import { sendRequest } from '../helpers/httpRequest'
 import { getProviderKeys } from '../lib/data'
 import { Provider, ServiceResponse } from '../types/general'
 import { ServiceKey } from '../types/admin'
+import fetch from 'node-fetch'
```
SahilKalra98: sendRequest results in unexpected end of JSON input in the response body, but the response goes through and sets the CNAME records

src/providers/goDaddy.ts
```diff
@@ -50,39 +51,55 @@ export const setRecord = async (
...
+    method: 'PUT',
+    headers: {
+      Authorization: `sso-key ${findKey('GD_Key')}:${findKey('GD_Secret')}`,
+      'Content-Type': 'application/json'
+    },
+    body: JSON.stringify(cnameData)
+  }
+  try {
+    // eslint-disable-next-line
+    const cnameResponse = await fetch(cnameURL, cnameOptions)
```
songz:
```
const results = await Promise.all([fetch(url, options), fetch(cnameUrl, cnameOptions)])
```

PR #176, opened 2019-09-21
Bug Fix - fixes ssl certPaths

SahilKalra98:


src/app.ts
```diff
@@ -55,16 +55,19 @@ const listener = (): void => {
 if (process.env.NODE_ENV === 'production') {
   const server = https.createServer(
     {
-      SNICallback: (domain, cb) => {
+      SNICallback: (host, cb) => {
         // escape characters required or readFileSync will not find file
         const homePath = process.env.HOME
+        /* eslint-disable */
+        const [subDomain, domain, topLevelDomain] = host.split('.')
```
songz: what if url was `blah1.blah2.blah3.albertow.com`

songz: subDomains can have multiple `.`

src/app.ts
```diff
@@ -55,16 +55,19 @@ const listener = (): void => {
...
+        const filteredDomain = `${domain}.${topLevelDomain}`
         const secureContext = tls.createSecureContext({
           /* eslint-disable */
           key: fs.readFileSync(
-            `${homePath}/\.acme\.sh/*\.${domain}/*\.${domain}\.key`
+            `${homePath}/\.acme\.sh/*\.${filteredDomain}/*\.${filteredDomain}\.key`
           ),
           cert: fs.readFileSync(
-            `${homePath}/\.acme\.sh/*\.${domain}/*\.${domain}\.cer`
+            `${homePath}/\.acme\.sh/*\.${filteredDomain}/*\.${filteredDomain}\.cer`
```
songz: for websockets and other requests to work, we should be using `fullChain.cer` I believe. Could you look into the `acme` folder to find out the fullpath of the domain name?

src/app.ts
```diff
@@ -55,16 +55,20 @@ const listener = (): void => {
...
         const homePath = process.env.HOME
+        /* eslint-disable */
+        const filteredHost = host.split('.')
+        const [domain, topLevelDomain] = filteredHost.slice(filteredHost.length-2, filteredHost.length)
+        const filteredDomain = `${domain}.${topLevelDomain}`
         const secureContext = tls.createSecureContext({
           /* eslint-disable */
           key: fs.readFileSync(
-            `${homePath}/\.acme\.sh/*\.${domain}/*\.${domain}\.key`
+            `${homePath}/\.acme\.sh/*\.${filteredDomain}/*\.${filteredDomain}\.key`
```
songz: this does not take care of the root certs.

PR #183, opened 2019-09-21
filtering mapped port and ip for proxy

allopez7:


src/app.ts
```diff
@@ -77,8 +77,16 @@ if (process.env.NODE_ENV === 'production') {
...
-      res.end('hello world')
+    const [totalDomain, mappings] = [(req.headers.host).split('.'), getMappings()]
+    const [domain, subdomain] = [totalDomain.slice(-2).join(''), totalDomain[0]]
+    let mappedDomain;
+    if(totalDomain.length === 3){
+    	mappedDomain = mappings.filter(e=>e.subDomain === subdomain && e.domain === domain)
     }
+	mappedDomain = domain
+	const {ip, port} = mappedDomain
+	console.log('ip', ip, 'port', port)
```
allopez7: removing

src/app.ts
```diff
@@ -77,8 +77,16 @@ if (process.env.NODE_ENV === 'production') {
       }
     },
     (req, res) => {
-      res.end('hello world')
+    const [totalDomain, mappings] = [(req.headers.host).split('.'), getMappings()]
+    const [domain, subdomain] = [totalDomain.slice(-2).join(''), totalDomain[0]]
+    let mappedDomain;
+    if(totalDomain.length === 3){
+    	mappedDomain = mappings.filter(e=>e.subDomain === subdomain && e.domain === domain)
```
allopez7: will add "find"

src/app.ts
```diff
@@ -77,8 +77,16 @@ if (process.env.NODE_ENV === 'production') {
...
     }
+	mappedDomain = domain
```
allopez7: domain is a string but should be an object. Will fix this issue.

src/app.ts
```diff
@@ -77,6 +80,24 @@ if (process.env.NODE_ENV === 'production') {
...
+      ]
+      let mappedDomain
+      if (totalDomain.length === 3) {
+        mappedDomain = mappings.find(
+          e => e.subDomain === subdomain && e.domain === domain
+        )
+      }
+      if(totalDomain.length === 2){
+      	mappedDomain = mappings.find(e=>e.domain === domain && !e.subDomain)
+      }
```
songz: All these lines of code can be condensed into 3 lines: 
```
const {ip, port} = mappings.find( ({subDomain, domain}) => {
  return `${subDomain}.${domain}` === req.headers.host
}) || {}
```

src/app.ts
```diff
@@ -87,7 +90,16 @@ if (process.env.NODE_ENV === 'production') {
...
     },
     (req, res) => {
-      res.end('hello world')
+      try {
+        const mappings = getMappings()
+        const { ip, port } =
+          mappings.find(({ subDomain, domain }) => {
+            return `${subDomain}.${domain}` === req.headers.host
+          }) || {}
+        if (port) return proxy.web(req, res, { target: `${ip}:${port}` })
```
songz: Please handle errors gracefully.
```
if (!port) return res.send('Cannot redirect')
return proxy.web....
```

songz: and then please add error handler. When there is an error, please do `res.send('Error communicating with server that runs <hostname>')`

src/app.ts
```diff
@@ -87,10 +92,31 @@ if (process.env.NODE_ENV === 'production') {
...
     (req, res) => {
-      res.end('hello world')
+      try {
+        const mappings = getMappings()
+        const { ip, port }: ProxyMapping =
+          mappings.find(({ subDomain, domain }) => {
+            return `${subDomain}.${domain}` === req.headers.host
+          }) || {}
+        if (!port) return res.end('Cannot redirect')
+        if (port)
```
songz: remove. unnecessary, because above you returned.

src/app.ts
```diff
@@ -87,10 +92,31 @@ if (process.env.NODE_ENV === 'production') {
...
     },
     (req, res) => {
-      res.end('hello world')
+      try {
+        const mappings = getMappings()
+        const { ip, port }: ProxyMapping =
+          mappings.find(({ subDomain, domain }) => {
+            return `${subDomain}.${domain}` === req.headers.host
+          }) || {}
+        if (!port) return res.end('Cannot redirect')
```
songz: You are not redirecting. `res.end('Not Found')`

songz: Also, `if (!port || !ip)`

src/app.ts
```diff
@@ -8,8 +8,15 @@ import { hashPass } from './helpers/crypto'
...
 import tls from 'tls'
-import { getAvailableDomains } from './lib/data'
+import { getAvailableDomains, getMappings } from './lib/data'
 import { isCorrectCredentials } from './auth'
+import httpProxy from 'http-proxy'
+import { ProxyMapping } from './types/general'
+
+const proxy = httpProxy.createProxyServer({})
+proxy.on('error', err => {
+  console.log('error', err)
```
songz: could you put in a better log message? 
```
console.error('proxy error', err)
```

src/app.ts
```diff
@@ -87,10 +92,25 @@ if (process.env.NODE_ENV === 'production') {
...
-      res.end('hello world')
+      try {
+        const mappings = getMappings()
+        const { ip, port }: ProxyMapping =
+          mappings.find(({ subDomain, domain }) => {
+            return `${subDomain}.${domain}` === req.headers.host
+          }) || {}
+        if (!port || !ip) return res.end('Not Found')
+        proxy.web(req, res, { target: `http://${ip}:${port}` }, err => {
+          console.log('Error communicating with server', err)
```
songz: `console.error`

src/app.ts
```diff
@@ -87,10 +92,25 @@ if (process.env.NODE_ENV === 'production') {
...
+          res.end(
+            `Error communicating with server that runs ${req.headers.host}`
+          )
+        })
+      } catch (err) {
+        console.log('Error: proxy failed', err)
```
songz: `console.error`

src/app.ts
```diff
@@ -8,8 +8,15 @@ import { hashPass } from './helpers/crypto'
...
 import tls from 'tls'
-import { getAvailableDomains } from './lib/data'
+import { getAvailableDomains, getMappings } from './lib/data'
 import { isCorrectCredentials } from './auth'
+import httpProxy from 'http-proxy'
+import { ProxyMapping } from './types/general'
+
+const proxy = httpProxy.createProxyServer({})
+proxy.on('error', err => {
+  console.log('Proxy error', err)
```
songz: `console.error`

PR #184, opened 2019-09-21
close #180 Port and IP should be Optional

### LINKED ISSUE #180: UI: Should say optional
Screen Shot 2019-09-20 at 10 28 06 PM

SahilKalra98:
image

### DIFF
--- a/views/client.ejs
+++ b/views/client.ejs
@@ -1,56 +1,55 @@
 <!DOCTYPE html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <title></title>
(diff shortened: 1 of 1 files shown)


views/client.ejs
```diff
@@ -31,11 +31,11 @@
       </div>
       <div class="form-group mx-sm-3 mb-2">
         <label for="port" class="sr-only">Port</label>
-        <input type="text" class="form-control port" id="port" placeholder="Port Number">
+        <input type="text" class="form-control port" id="port" placeholder="Port Number (Optional)">
```
songz: they should be at the `label`. Port number should be an example of what the user is supposed to put in.

PR #187, opened 2019-09-21
fix #186 refactor Automatic Generation of Ports

### LINKED ISSUE #186: Refactor Auto Generation of Ports

rkalra247:


src/api/mapping.ts
```diff
@@ -11,13 +11,12 @@ const exec = util.promisify(cp.exec)
...
-  if(req.body.port === ''){
-    sortedKeys.forEach((e) => {
-      if(e.port === portCounter.toString()) portCounter += 1
-    })
-  }
+  domainKeys.reduce((acc, e) =>{
+    acc[e.port] = true
+    if(acc[portCounter]) portCounter += 1
+    return acc
+  }, {})
```
```
const getNextPort=(map, start=3002) => {
...
}
...
const map = domainKeys.reduce.....
portCounter = getNextPort(map)
```

src/api/mapping.ts
```diff
@@ -11,13 +11,16 @@ const exec = util.promisify(cp.exec)
 
 mappingRouter.post('/', (req, res) => {
   const domainKeys = getMappings()
-  const sortedKeys = domainKeys.sort((a,b) =>parseInt(a.port) - parseInt(b.port))
-  let portCounter = 3002
-  if(req.body.port === ''){
-    sortedKeys.forEach((e) => {
-      if(e.port === portCounter.toString()) portCounter += 1
-    })
+  const getNextPort = (map, start=3002):number =>{
```
songz: this helper function does not need closure (therefore, does not need to be defined inside the callback handler). please move outside.

PR #191, opened 2019-09-22
fix #189 only allow port values > 3001 to be registered

### LINKED ISSUE #189: Only allow port values > 3001 to be registered

rkalra247:


src/public/client.ts
```diff
@@ -161,7 +161,8 @@ create.onclick = (): void => {
   const subDomain = helper.getElement('.subDomain') as HTMLInputElement
   const port = helper.getElement('.port') as HTMLInputElement
   const ipAddress = helper.getElement('.ipAddress') as HTMLInputElement
-
+  if (parseInt(port.value) < 3001)
+    return alert('Please enter value for port > 3001')
```
songz: Please add a server side check as well.

src/api/mapping.ts
```diff
@@ -16,6 +16,9 @@ const getNextPort = (map, start=3002):number =>{
 
 mappingRouter.post('/', (req, res) => {
   const domainKeys = getMappings()
+  if(parseInt(req.body.port) < 3001){
+    return res.status(400)
```
```
return res.status(400).json({message: "Port cannot be smaller than 3001"})
```

songz: Lastly, when you receive this response from the user, could you fire off an alert?

PR #198, opened 2019-09-23
reroute http to https 

allopez7:


src/app.ts
```diff
@@ -90,8 +90,13 @@ if (process.env.NODE_ENV === 'production') {
       res.end('hello world')
     }
   )
-
   server.listen(443)
+
+  const httpApp = express()
+  httpApp.get('/*', (req, res) => {
+    res.redirect(`https://${req.headers.host}${req.path}`)
```
songz: You are not handling query parameters.

src/app.ts
```diff
@@ -112,6 +112,13 @@ if (process.env.NODE_ENV === 'production') {
     }
   )
   server.listen(443)
+
+  const httpApp = express()
+  httpApp.get('/*', (req, res) => {
+    const params = req.query ? `?${req.headers.host.split('?')[1]}` : ''
```
songz: if for whatever reason.... there is a `?` in the query params, this might fail. I think this is better:
```
req.headers.host.split('?').shift().join('?')
``

Comment (songz) [reply src/app.ts]: never mind.

File: src/app.ts [hunk 3]
```diff
@@ -112,6 +112,13 @@ if (process.env.NODE_ENV === 'production') {
...
+    res.redirect(`https://${req.headers.host}${req.path}/${params}`)
```
songz: 1. question: Do you need `/` between host and path, or does path already prefixed by `/`?
2. query perimeters: you need to add `?` right?

PR #214, opened 2019-09-24
fix undefined parameter bug

allopez7:


src/server/server.ts
```diff
@@ -111,9 +111,7 @@ const startProxyServer = (homePath: string): void => {
 
   const httpApp = express()
   httpApp.get('/*', (req, res) => {
-    const params = Object.keys(req.query).length
-      ? `?${req.headers.host.split('?')[1]}`
-      : ''
+    const params = `?${req.headers.host.split('?')[1]}` || ''
```
allopez7: still incorrect, will fix.

PR #225, opened 2019-09-27
Integration test setup, fix security issue

songz:
Currently, password check is literally useless. lol.

### DIFF
--- a/src/tests/api.test.ts
+++ b/src/tests/api.test.ts
@@ -16,7 +16,7 @@ describe('/api', () => {
     server.close()
   })
 
-  it('Should respond with 403 if pw does not match', async () => {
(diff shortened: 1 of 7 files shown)


src/tests/api.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  beforeAll(async () => {
+    server = await startAppServer(TEST_PORT, ADMIN)
+  })
+
+  afterAll(() => {
+    server.close()
+  })
+
+  it('Should respond with 403 if pw does not match', async () => {
```
allopez7: 403 should say 401

songz: Good catch!

Issue #229, opened 2019-09-28
ISSUES before launch


songz:
~~Download production config~~
~~login password -> input box cannot be plain text.~~
reconfigure -> PATCH
CSS (input box width)
EDIT mapping doesn't work with git
Delete mapping should delete repository as well.

songz:
~~Download production config~~
~~login password -> input box cannot be plain text.~~
~~reconfigure -> PATCH~~
~~CSS (input box width)~~
EDIT mapping doesn't work with git
Delete mapping should delete repository as well.

songz:
All done, thanks @SahilKalra98 @albertoelopez @rkalra247 @Wolfy64

PR #240, opened 2019-10-04
Edit mapping doesn't work with git

SahilKalra98:
This is issue 5 from #229


src/api/mapping.ts
```diff
@@ -106,7 +106,37 @@ mappingRouter.patch('/edit/:id', (req, res) => {
   const updatedDomain = domains.find(
     (element: Mapping) => element.id === req.params.id
   )
-  res.json(updatedDomain)
+  const updatedConfig = {
```
songz: PR 1: helper functions.
PR 2: implement this by calling helper functions.

PR #243, opened 2019-10-07
fix# 242 Integration Test for POST/mappings

rkalra247:


src/tests/integration/mapping.test.ts
```diff
@@ -0,0 +1,36 @@
...
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({
+        domain: 'Rahul',
+        subdomain: 'testing',
+        port: '5678'
+
+      })
+    }).then(r => r.json()).then((data)=>{
+      expect(data.port).toEqual('5678')
```
songz: `data` has more fields than just `port`. Could you test all of the fields?

src/tests/integration/mapping.test.ts
```diff
@@ -0,0 +1,38 @@
...
+  it('checks mappings for newly added mapping', async()=>{
+    await fetch(`${apiURL}/api/mappings`, {
+      method: 'POST', 
+      headers: {
+        authorization: ADMIN,
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({
+        domain: 'Rahul',
+        subDomain: 'testing',
```
songz: hmmm.... creating 2 same subDomains seems to be a problem, creating an issue to track this:
https://github.com/garageScript/myProxy/issues/244

For now, I think you need a variable to guarantee uniqueness for every time this test runs. 
```
const subDomain = `testing${uuidv4()}`
...
    body: JSON.stringify({
      domain...
      subDomain....
     })
....
```

src/tests/integration/mapping.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  beforeAll(async ()=>{
+    server = await startAppServer(TEST_PORT, ADMIN)
+  })
+
+  afterAll(()=>{
+    server.close()
+  })
+
+
+  const subDomain = `testing${uuidv4()}`
```
songz: put this inside `it`. Every test might be creating a new subdomain.

src/tests/integration/mapping.test.ts
```diff
@@ -0,0 +1,43 @@
...
+        domain: 'Rahul',
+        subDomain,
+        port: '5678'
+
+      })
+    }).then(r => r.json()).then((data)=>{
+      expect(data.port).toEqual('5678')
+      expect(data.subDomain).toEqual(subDomain)
+      expect(data.domain).toEqual('Rahul')
+      expect(data.fullDomain).toEqual(`${subDomain}.Rahul`)
```
songz: could you make a `domain` and a `port` variable just like `subDomain`? This way, if we need to change anything we just change the value in the variable instead of changing it in multiple places.

songz: `${subDomain}.${domain}`

PR #245, opened 2019-10-08
Integration test for api/availableDomains

Wolfy64:


src/tests/api.test.ts
```diff
@@ -35,4 +35,25 @@ describe('/api', () => {
...
   })
+
+  describe('api/availableDomains', () => {
+    it('Should respond with 200', async () => {
+      const response = await fetch(`${apiUrl}/api/availableDomains`,{
+        headers: {
+          authorization: ADMIN,
+        },
+      })
+      expect(response.status).toEqual(200)
```
songz: instead of having another instance test `it`, just add another expect here:
```
const data = await response.json()
      expect(data).toEqual([])
```

PR #246, opened 2019-10-08
Integration test: Delete mapping

allopez7:


src/tests/integration/mapping.test.ts
```diff
@@ -2,8 +2,8 @@ import { startAppServer } from '../../server/server'
 import fetch from 'node-fetch'
 import uuidv4 from 'uuid/v4' 
 
-const TEST_PORT = process.env.PORT || 50604
```
allopez7: will change these back

src/tests/integration/mapping.test.ts
```diff
@@ -42,4 +40,39 @@ describe('/api', () =>{
       expect(data.fullDomain).toEqual(`${subDomain}.${domain}`)
     })
   })
+
+  it('Delete mapping', async()=>{
+    const subDomain = `delete${uuidv4()}`
+    const domain = 'albertow'
+    const port = '4500'
+    await fetch(`${apiURL}/api/mappings`, {
```
allopez7: will refactor using Promise.all soon

src/tests/integration/mapping.test.ts
```diff
@@ -42,4 +40,39 @@ describe('/api', () =>{
...
+      return data
+    }).then(async (data)=>{
+      await fetch(`${apiURL}/api/mappings/delete/${data.id}`, {
+        method: 'DELETE', 
+        headers: {
+          authorization: ADMIN,
+          'Content-Type': 'application/json'
+        }
+      }).then(r=>r.json()).then((del)=>{
+        expect(del.id).toEqual(data.id)
```
allopez7: adding more statements

src/tests/integration/mapping.test.ts
```diff
@@ -42,4 +40,39 @@ describe('/api', () =>{
...
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({
+        domain,
+        subDomain,
+        port,
+
+      })
+    }).then(r => r.json()).then((data)=>{
+      expect(data.port).toEqual(port)
```
allopez7: will delete these since Rahul already tested post request.

src/tests/integration/mapping.test.ts
```diff
@@ -42,4 +40,36 @@ describe('/api', () =>{
       expect(data.fullDomain).toEqual(`${subDomain}.${domain}`)
     })
   })
+
+  it('Delete mapping', async()=>{
+    const subDomain = `delete${uuidv4()}`
+    const domain = 'albertow'
+    const port = '4500'
+    await fetch(`${apiURL}/api/mappings`, {
```
songz: let's create a helper function, there's alot of duplicate code.
```
const mappingAdapter = (path='/', method, body) => {
  const options = {
    method, headers: {
    ....
    }
  }
  if (body) {
    options.body = JSON.stringify(body)
  }
  return fetch(`${apiURL}/api/mappings${path}`, options)
}
```

Then to use this:
```
await mappingAdapter('/', 'POST', {domain, subDomain, port})
const deleteResponse = await mappingAdapter(`/${data.id}`, 'POST', {domain, subDomain, port})
const deleteResult = await deleteResponse.json()
expect(deleteResult.port).toEqual(port)
.....
```

src/tests/integration/mapping.test.ts
```diff
@@ -42,4 +40,36 @@ describe('/api', () =>{
...
+        authorization: ADMIN,
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({
+        domain,
+        subDomain,
+        port,
+      })
+    }).then(r => r.json()).then(async (data)=>{
+      await fetch(`${apiURL}/api/mappings/delete/${data.id}`, {
```
songz: path is wrong. should be `${apiURL}/api/mappings/${data.id}`. Is this a bug with the codebase?

allopez7: yes, that is the end point on the server

allopez7: del

allopez7: Will refactor but nearly every other end point has the same issue. I will post an issue to refactor.

songz: https://github.com/garageScript/myProxy/pull/246/files

src/tests/integration/mapping.test.ts
songz: Should create a refactor story.

songz: you or @rkalra247 can pick this one up.

src/tests/integration/mapping.test.ts
```diff
@@ -42,4 +40,36 @@ describe('/api', () =>{
...
+          authorization: ADMIN,
+          'Content-Type': 'application/json'
+        }
+      }).then(r=>r.json()).then((del)=>{
+        expect(del.port).toEqual(port)
+        expect(del.subDomain).toEqual(subDomain)
+        expect(del.domain).toEqual(domain)
+        expect(del.fullDomain).toEqual(`${subDomain}.${domain}`)
+        expect(del.id).toEqual(data.id)
+      })
```
songz: could you do another fetch after this.... to ` GET /api/mappings/${data.id}` and make sure the response is null? This way we make sure its 100% deleted. Sometimes, just because `delete` sends back a correct response, doesn't mean it actually got deleted.

src/tests/integration/mapping.test.ts
songz: Can be marked as resolved for now.

allopez7: I'm already done, just need to push up.

src/types/tests.ts
```diff
@@ -0,0 +1,12 @@
+type Options = {
+  headers: unknown
```
allopez7: typescript on push is converting my headers any to unknown.

src/tests/helpers/mappingAdapter.ts
```diff
@@ -0,0 +1,23 @@
+const TEST_PORT = process.env.PORT || 50604
```
allopez7: I wasn't sure if these should be passed into the function or stated here. Passing them into the function would be sort of painful for the developer.

songz: this is okay. the user simply just do ```PORT=3095 npm run test``` to run the test if they want to run on different port.

src/tests/integration/mapping.test.ts
```diff
@@ -42,6 +43,30 @@ describe('/api', () => {
...
+    const delMapping = await mappingAdapter(`/delete/${mapping.id}`, 'DELETE')
+    expect(delMapping.status).toEqual(200)
+    const deletedMapping = await delMapping.json()
+    const getDeletion = await mappingAdapter(`/${mapping.id}`, 'GET')
+    const checkDeletion = await getDeletion.json()
+    expect(deletedMapping.port).toEqual(port)
+    expect(deletedMapping.subDomain).toEqual(subDomain)
+    expect(deletedMapping.domain).toEqual(domain)
+    expect(deletedMapping.fullDomain).toEqual(`${subDomain}.${domain}`)
+    expect(deletedMapping.id).toEqual(mapping.id)
```
songz: move these above `getDeletion`. If any of these fails, we don't need to get the mapping

src/tests/integration/mapping.test.ts
```diff
@@ -42,6 +43,30 @@ describe('/api', () => {
...
+      domain,
+      subDomain,
+      port,
+    })
+    expect(createMapping.status).toEqual(200)
+    const mapping = await createMapping.json()
+    const delMapping = await mappingAdapter(`/delete/${mapping.id}`, 'DELETE')
+    expect(delMapping.status).toEqual(200)
+    const deletedMapping = await delMapping.json()
+    const getDeletion = await mappingAdapter(`/${mapping.id}`, 'GET')
```
songz: `getMapping` -> This API is to get a mapping. getting deletion doesn't make sense.

src/api/mapping.ts
```diff
@@ -97,6 +97,12 @@ mappingRouter.get('/', (req, res) => {
   res.json(domains)
 })
 
+mappingRouter.get('/:id', (req, res) => {
+  const domains = getMappings()
+  const foundDomain = domains.find(e => e.id === req.params.id)
+  res.json({ checkDomain: foundDomain })
```
songz: `checkDomain` is not following REST Instead:
```
res.json(foundDomain || {})
```

src/tests/integration/mapping.test.ts
```diff
@@ -42,6 +43,31 @@ describe('/api', () => {
...
+    expect(delMapping.status).toEqual(200)
+    const deletedMapping = await delMapping.json()
+    expect(deletedMapping.port).toEqual(port)
+    expect(deletedMapping.subDomain).toEqual(subDomain)
+    expect(deletedMapping.domain).toEqual(domain)
+    expect(deletedMapping.fullDomain).toEqual(`${subDomain}.${domain}`)
+    expect(deletedMapping.id).toEqual(mapping.id)
+    const getMapping = await mappingAdapter(`/${mapping.id}`, 'GET')
+    expect(getMapping.status).toEqual(200)
+    const checkDeletion = await getMapping.json()
```
songz: better name: `mappingData`

src/tests/integration/mapping.test.ts
```diff
@@ -42,6 +43,31 @@ describe('/api', () => {
...
+    expect(checkDeletion.checkDomain).toEqual(undefined)
```
songz: `checkDomain` is not restful.
```
expect(Object.keys(checkDomain).length).toEqual(0)
```

PR #247, opened 2019-10-08
Test /api/admin/sslCerts

Wolfy64:


src/tests/integration/admin.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    })
+
+    const request = await sslCertsMock()
+    expect(sslCertsMock).toHaveBeenCalledTimes(1)
+    expect(request).toEqual({
+      success: true,
+      message: 'SSL Certs and domain name records successfully created'
+    })
+  })
+})
```
Wolfy64: I'm not really sure how to test this api...

src/tests/api.test.ts
```diff
@@ -1,7 +1,7 @@
 import { startAppServer } from '../server/server'
 import fetch from 'node-fetch'
 
-const TEST_PORT = process.env.PORT || 4998
+const TEST_PORT = 4242
```
songz: sometimes, when I run tests, I run it on a different port. I have apps running on 4242 on my local machine, so `process.env.PORT` must remain.

src/tests/integration/admin.test.ts
songz: haha you are testing to see if `mockResorvedValue` works. I don't think there's a way to test `/sslCerts` path.

src/tests/api.test.ts
Wolfy64: I got it. I removed it because it circleCI was failing... 🧐

songz: hmmmm.... that sucks.
maybe `process.env.PORT || 4242`?

songz: or `process.env.TEST_PORT || 4242`

PR #248, opened 2019-10-08
fix #244 Mapping should not allow same subDomains for the same domain

### LINKED ISSUE #244: Mapping should not allow same subDomains for the same domain
**Describe the bug**
If I create a `hello` subdomain for `c0d3.com` domain already, nobody else should be able to create a `hello` subdomain for `c0d3.com` domain. It will be confusing.

Bug in the API

rkalra247:


src/api/mapping.ts
```diff
@@ -45,6 +45,13 @@ mappingRouter.post('/', (req, res) => {
   const scriptPath = '.scripts'
 
   const respond = () => {
+    const verifySubDomain = domainKeys.find(
```
songz: this check should not be in the `respond` function, it might cause errors in the future. in line 19 there's already a check. Just move this to around the same area, so all the checks stay around the same area.

src/api/mapping.ts
```diff
@@ -18,6 +18,13 @@ mappingRouter.post('/', (req, res) => {
   if (parseInt(req.body.port) < 3001) {
     return res.status(400).json({ message: 'Port cannot be smaller than 3001' })
   }
+  const verifySubDomain = domainKeys.find(
```
songz: rename `verifySubdomain` to `existingSubDomain`

src/tests/integration/mapping.test.ts
```diff
@@ -1,45 +1,75 @@
...
+      .then(data => {
+        expect(data.port).toEqual(port)
+        expect(data.subDomain).toEqual(subDomain)
+        expect(data.domain).toEqual(domain)
+        expect(data.fullDomain).toEqual(`${subDomain}.${domain}`)
+      })
+  })
 
+  it('checks no duplicate subdomain is created for same domain', async () => {
+    const subDomain = 'testingSubDomain'
```
songz: should have `uuidv4()`

PR #249, opened 2019-10-08
Edit mappings doen't work with git

SahilKalra98:
~~Need to test this still~~


src/api/mapping.ts
```diff
@@ -31,26 +33,14 @@ mappingRouter.post('/', (req, res) => {
...
   const fullDomain = `${req.body.subDomain}.${req.body.domain}`
+  const prodConfigApp = [...prodConfigure.apps][0]
+  prodConfigApp.name = fullDomain
+  prodConfigApp.env_production.PORT = parseInt(req.body.port || portCounter, 10)
   const prodConfig = {
-    apps: [
-      {
-        name: fullDomain,
-        script: 'npm',
-        args: 'start',
```
songz: we need the 2 above for running React Apps.

PR #252, opened 2019-10-09
Quick fix

Wolfy64:


src/auth/index.ts
```diff
@@ -9,15 +9,13 @@ const isCorrectCredentials = (password: string, correct: string): boolean => {
...
-      !isCorrectCredentials(
-        (req.headers.authorization as string) || '',
-        password
-      )
-    ) {
-      return res.status(401).send('Unauthorized')
+    const { authorization = '' } = req.headers
+
+    if (authorization) {
+      const isAdmin = !isCorrectCredentials(authorization as string, password)
```
songz: `isAdmin = isCorrect....`

src/auth/index.ts
```diff
@@ -9,15 +9,13 @@ const isCorrectCredentials = (password: string, correct: string): boolean => {
...
+      if (!adminPass && isAdmin) return res.status(401).send('Unauthorized')
```
songz: `if (!adminPass && !isAdmin)...`

PR #259, opened 2019-10-15
Test Fix: Deletes Mapping after creation/Avoid dummy data in DB

SahilKalra98:


src/tests/integration/mapping.test.ts
```diff
@@ -22,25 +22,22 @@ describe('/api', () => {
...
-      body: JSON.stringify({
-        domain,
-        subDomain,
-        port,
-      }),
+    const postResponse = await mappingAdapter('/', 'POST', {
+      domain,
+      subDomain,
+      port,
     })
```
songz: For thhe future, you can diectly do:
```
const postMapping = await mappingAdapte(....).thhen(r => r.json())
```

Cuently you are not using `postResponse`, so it doesn't have to be there.

PR #268, opened 2019-10-26
Fix Invalid requests causes server to crash

### LINKED ISSUE #263: Invalid requests causes server to crash
**Describe the bug**
If a request comes to server and server does not have the certificate, the server crashes.
After the server crashes, `pm2` will restart it.

However, this introduces a security issue. Somebody can call the server repeatedly with a bad request and crash the proxy repeatedly.

**To Reproduce**
Start the server. Access the server directly via the server IP address. myProxy should crash.

**Expected behavior**
myProxy should not crash. As a proxy, it should never crrashh.

**Screenshots**
Server logs:
Screen Shot 2019-10-20 at 1 58 41 PM

**Additional context**
Issue is probably a lack of error catching here: https://github.com/garageScript/myProxy/blob/master/src/server/server.ts#L88

Wolfy64:
close #263


src/helpers/SNICallback.ts
```diff
@@ -0,0 +1,32 @@
...
+const readFileAsync = util.promisify(fs.readFile)
+
+const readFile = async (path: string) => {
+  const data = await readFileAsync(path, 'utf8')
+  return data || null
+}
+
+const SNICallback = async (host, cb) => {
+  // TLD -> Top-Level Domain | SLD -> Second-Level Domain
+  const [TLD, SLD, ...subDomains] = host.split('.').reverse()
```
songz: very smart

PR #269, opened 2019-10-28
run autofix

Wolfy64:


Wolfy64:
This PR do not any change except code format with prettier by running `npm autofix`

PR #270, opened 2019-10-28
Fix prettier check

Wolfy64:
Prettier:check was not working! Now CircleCI should fail if the code don't match Prettier style.
**Don't forget to kick `npm run autofix`  before each PR** 😁


src/api/mapping.ts
```diff
@@ -136,15 +136,17 @@ mappingRouter.patch('/edit/:id', async (req, res) => {
...
       cd ${projectPath}/${updatedDomain.fullDomain}
       echo 'module.exports = ${JSON.stringify(
-    updatedConfig
-  )}' > deploy.config.js
+        updatedConfig
+      )}' > deploy.config.js
       git add .
       git commit -m "Edits deploy config file"
       `,
+    /*eslint-enable */
```
Wolfy64: This line was in conflict between `prettier:check` & `eslint:check`

PR #275, opened 2019-10-30
Setup Script Fix

hwong0305:
- Added changes that were generated by npm run autofix.
- Update Setup.SH bash file to install Node Version 12
- Updated Readme and Package.json to include this reference.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,6 @@
     "test": "jest --coverage",
     "test:dev": "jest --coverage --watch",
     "start": "tsc -p . --watch & nodemon build/app.js",
-    "setup": "./scripts/setup.sh",
     "build": "tsc -p .",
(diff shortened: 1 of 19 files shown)


package.json
```diff
@@ -7,7 +7,6 @@
     "test": "jest --coverage",
     "test:dev": "jest --coverage --watch",
     "start": "tsc -p . --watch & nodemon build/app.js",
-    "setup": "./scripts/setup.sh",
```
Wolfy64: Why did you removed `npm run setup` ?

songz: I actually prefer it removed. Let's not have scripts that we don't run. In a real `setup` for new users, the script actually installs `nodejs` if it doesn't exist. This would be impossible to do if we force users to `npm run setup` because npm wouldn't exist.

PR #284, opened 2019-11-02
update fix #274 Button to redirect to admin page

### LINKED ISSUE #274: It would be useful if there's an added button that redirects the user from the admin page to main page
It would be useful if there's an added button that redirects the user from the admin page to main page

rkalra247:


views/admin/providers.ejs
```diff
@@ -2,7 +2,15 @@
 <%- include('../layout/head.ejs') %>
 
   <div class="container" style="margin-top: 50px;">
+    <div class=headContainer style="display:flex;">
     <h2>Host Keys</h2>
+    <a
```
songz: `a` links needs a `href` attribute, to tell the link where to go to.

PR #290, opened 2019-11-08
 fix #287 sets up get endpoint for sshKeys

### LINKED ISSUE #287: Ability to add / remove ssh keys
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I have to deploy for other people . They should be able to deploy their apps themselves.

**Describe the solution you'd like**
ability for admins to add / remove other people's ssh keys

**Additional context**

rkalra247:


src/api/sshKeys.ts
```diff
@@ -0,0 +1,14 @@
+import express from 'express'
+import fs from 'fs'
+const sshKeyRouter = express.Router()
+
+sshKeyRouter.get('/', (req, res) => {
+  //res.json([])
+  fs.readFile('/home/noob101/.ssh/authorized_keys', (error, data) => {
```
songz:
```
const os = require('os');

console.log(os.homedir());
```
https://stackoverflow.com/questions/9080085/node-js-find-home-directory-in-platform-agnostic-way

PR #301, opened 2019-11-26
allow root domain to be proxied correctly

rkalra247:


songz:
Good job finding the bug!

PR #304, opened 2019-12-01
Implement Icons for Server Status

hwong0305:
The server will display the following:
- No Icon if it's just created
- Red Icon if server is stopped or errored out
- Green Icon if server is running

Includes the following code fixes:
-  Users cannot use the same subdomain for different domains. 
-  Client doesn't show correct error message if subdomain already exists.


src/api/mapping.ts
```diff
@@ -23,13 +23,13 @@ mappingRouter.post('/', async (req, res) => {
   if (parseInt(req.body.port) < 3001) {
     return res.status(400).json({ message: 'Port cannot be smaller than 3001' })
   }
-  const existingSubDomain = domainKeys.find(
-    e => e.subDomain === req.body.subDomain
-  )
+  const existingSubDomain =
+    domainKeys.find(e => e.subDomain === req.body.subDomain) &&
+    domainKeys.find(e => e.domain === req.body.domain)
```
songz: since you discovered a bug, could you add a test to make sure nobody regresses on thhis bug inn the future?

songz: Ideally, this bug fix and your `status` feature, should have been 2 separate PRs. In this case, since its already done, its okay. Would need a test for this fix, thank you.

songz: what about finding `fullDomains` instead.... does that work?

hwong0305: I wrote the test for it. Let me see if i can figure out how to do it with one check.

src/api/mapping.ts
```diff
@@ -23,19 +23,17 @@ mappingRouter.post('/', async (req, res) => {
   if (parseInt(req.body.port) < 3001) {
     return res.status(400).json({ message: 'Port cannot be smaller than 3001' })
   }
-  const existingSubDomain = domainKeys.find(
-    e => e.subDomain === req.body.subDomain
-  )
+  let fullDomain = `${req.body.subDomain}.${req.body.domain}`
```
songz: What if its the root domain? `blah.com`

You need the code below: 
```
if (!req.body.subDomain) {
   fullDomain = ....
}
```

src/public/client.ts
```diff
@@ -220,3 +223,27 @@ fetch('/api/availableDomains')
...
+fetch('/api/status')
+  .then(r => r.json())
+  .then(data => {
+    const dataObj = JSON.parse(data.stdout).map(el => ({
+      name: el.name,
+      status: el.pm2_env.status
+    }))
+
+    dataObj.forEach(el => {
+      const statusContainer = document.getElementById(el.name)
```
songz: let's make this more modular by putting the status logic inside mapping item. Current implementation feels a little hacky and becomes difficult to maintain. Logically, MappingItem class should take care of everything relating to mapping (including status). To do this, change the code above from:
```
fetch('/api/mappings')
 .then(r => r.json())
 .then((data: Mapping[]) => {....
```

To:
```
Promise.all([
  fetch('/api/mappings').then( r => r.json()),
  fetch('/api/statuses').then( r => r.json())
]).then( ([mappings: Mapping[], statuses: Status[]]) => {
  const fullDomainStatusMap = statuses.reduce(....., {})
  domainList.innerHTML = ''
  mappings
    .reverse()
    .filter(e => e.domain && e.port && e.id && e.ip && e.gitLink && e.fullDomain)
   .forEach( mapping => {
      new MappingItem(mapping, fullDomainStatusMap[ mapping.fullDomain ] )
  })
})
```

Then update `MappingItem` class to draw directly into render (you will no longer need `id` for tag attribute).

src/public/client.ts
```diff
@@ -220,3 +223,27 @@ fetch('/api/availableDomains')
     selectedHost = domains[0]
     hostSelector.innerText = domains[0]
   })
+
+fetch('/api/status')
```
songz: REST convention - resource should be plural, `statuses`

src/tests/integration/mapping.test.ts
```diff
@@ -162,4 +162,48 @@ describe('/api', () => {
     const mappingData = await getMapping.json()
     expect(Object.keys(mappingData).length).toEqual(0)
   })
+
+  it('checks same subdomain can be used for different domains', async () => {
+    const subDomain = `testing${uuidv4()}`
+    const domain = 'VinDiesel'
+    const secondDomain = 'PaulWalker'
```
songz: declare variables close to where it gets used. In this case, `secondDomain` should be moved a few lines down, in the second `await mappingAdapter`

src/api/index.ts
```diff
@@ -2,13 +2,15 @@ import express from 'express'
...
 import sshKeyRouter from './sshKeys'
+import statusRouter from './serversStatus'
 import { getAvailableDomains } from '../lib/data'
 
 const apiRouter = express.Router()
 
 apiRouter.use('/admin', adminRouter.app)
 apiRouter.use('/mappings', mappingRouter)
 apiRouter.use('/sshKeys', sshKeyRouter)
+apiRouter.use('/status', statusRouter)
```
songz: REST convention - resource should be plural, `statuses`

src/api/serversStatus.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const statusRouter = express.Router()
+const exec = util.promisify(cp.exec)
+const { isProduction } = environment
+
+statusRouter.get('/', async (req, res) => {
+  try {
+    if (isProduction()) {
+      const data = await exec('su - myproxy -c "pm2 jlist"')
+      res.json(data)
```
songz: could you massage the data so the data output is predictable and easy to follow? Current code, new devs has no idea what `data` shape looks like. Instead:
```
const result = data.map(.....)
res.json(result)
```

src/public/client.ts
```diff
@@ -41,7 +41,7 @@ class MappingItem {
     domainList.appendChild(mappingElement)
     mappingElement.innerHTML = `
       <div style='width: 100%'>
-        <div style='display: flex'>
+        <div style='display: flex' id="${data.fullDomain}">
```
songz: look at comments above, we should not need `id=....` attribute.

src/api/mapping.ts
```diff
@@ -23,22 +23,19 @@ mappingRouter.post('/', async (req, res) => {
   if (parseInt(req.body.port) < 3001) {
     return res.status(400).json({ message: 'Port cannot be smaller than 3001' })
   }
-  const existingSubDomain = domainKeys.find(
-    e => e.subDomain === req.body.subDomain
-  )
+  const fullDomain = !req.body.subDomain
```
songz: avoid negating in an if-else
```
fullDomain = req.body.subDomain
  ? `${req.body.subDomain}.${....}`
  : `${req.body.domain}
```

src/public/client.ts
```diff
@@ -168,19 +201,32 @@ class MappingItem {
...
+      }
+    ]
+  }, [])
+
+  domainList.innerHTML = ''
+  fullDomainStatusMap.reverse()
+  fullDomainStatusMap
+    .filter(e => e.domain && e.port && e.id && e.gitLink && e.fullDomain)
+    .forEach(e => {
+      new MappingItem(e)
```
songz: Could this accomplish the same thing, without as much code?

* remove `const fullDomainStatusMap = ...` and instead use `mappings.reverse().filter....`
```
domainList.innerHTML = ''
domainList
  .reverse()
  .filter(....)
  .forEach( e => {
    e.status = statuses[e.fullDomain]
    new MappingItem(e
  })
```

src/api/serversStatus.ts
```diff
@@ -0,0 +1,33 @@
+import cp from 'child_process'
```
songz: could you add a simple integration test case for this? That sending a requset to `/api/serverStatus` to get back `{}`. This way, we at least test that this endpoint must be present and active.

Also, I like this filename as `serverStatus` better than the plural server from as it is now.

src/tests/integration/mapping.test.ts
```diff
@@ -162,4 +162,48 @@ describe('/api', () => {
...
+    expect(getMapping.status).toEqual(200)
+    const mappingData = await getMapping.json()
+    const getSecondMapping = await mappingAdapter(
+      `/${secondPostMapping.id}`,
+      'GET'
+    )
+    expect(getSecondMapping.status).toEqual(200)
+    const secondMappingData = await getSecondMapping.json()
+    expect(Object.keys(mappingData).length).toEqual(0)
+    expect(Object.keys(secondMappingData).length).toEqual(0)
```
songz: this test is confusing, because it is supposed to test... `check same subdomain can be used for different domains`
tests should be focused on exactly that. I think the most straightforward test to do is:

1. create 1 domain (no need to check response, because other tests took care of domains can be created)
2. create 2nd domain (different domain, same subdomain), check response (should be okay)
3. get all the mappings, expect 1, 2 to be in there
4. delete the mappings

src/api/serversStatus.ts
songz: Nvm about the simple integration test case, its already in the PR. Just need to change file name.

Issue #306, opened 2019-12-03
Cache lookup - Performance Improvement


songz:
Instead of looping over each array when finding a port, please create a caching mechanism (object) to do a O(1) lookup instead of the existing O(n) lookup. 

Especially for SNI Callback (https)
and Proxy redirection

https://github.com/garageScript/myProxy/blob/master/src/server/server.ts#L96

songz:
Blocking https://apps.c0d3.com/ upgrade atm.

songz:
PART 2: Cache SNICallback for `readFile` for keys and certificates. These should live in memory so we don't have to read from file for EVERY https certificate exchange.

PR #319, opened 2019-12-07
Implement Caching for Mapping

hwong0305:
#306


src/api/mapping.ts
```diff
@@ -93,11 +94,11 @@ mappingRouter.get('/', (req, res) => {
...
-  const deletedDomain = domains.find(e => e.id === req.params.id)
-  const updatedDomains = domains.filter(e => {
-    return e.id !== req.params.id
-  })
-  setData('mappings', updatedDomains)
+  const deletedDomain = getMappingFromId(req.params.id)
+  const updatedDomains = { ...domains }
+  delete updatedDomains[deletedDomain.fullDomain]
+  // Setting data as array to retain data structure of existing sites
+  setData('mappings', Object.values(updatedDomains))
```
hwong0305: If this line is not done, it will be incompatible with already launched myproxy servers after upgrading.  

When I was testing on my droplet, the create and delete servers functionality did not work correctly since the mappings object in data.db is not in the proper format.

src/lib/data.ts
```diff
@@ -1,13 +1,17 @@
...
-import { Mapping, Domain } from '../types/general'
+import { Mapping, MappingObj, Domain } from '../types/general'
 
 const data: DB = {
   serviceKeys: [],
   mappings: [],
   availableDomains: []
 }
 
+let mappingsCache: MappingObj | {} = {}
```
songz: `domainToMapping` and `idToMapping` are clearer variable names

src/lib/data.ts
```diff
@@ -19,6 +23,22 @@ fs.readFile('./data.db', (err, file) => {
   data.serviceKeys = fileData.serviceKeys || []
   data.mappings = fileData.mappings || []
   data.availableDomains = fileData.availableDomains || []
+
+  mappingsCache = data.mappings.reduce(
```
songz: could you create a helper functionn for this? I would call thhis function:
```
updateMappingCache( data.mappings )
```

src/lib/data.ts
```diff
@@ -29,6 +49,24 @@ const getData = (table: string): unknown => {
...
+      }),
+      {}
+    )
+    mappingsDict = initialData.reduce(
+      (obj, item) => ({
+        ...obj,
+        [item.id]: item
+      }),
+      {}
+    )
```
songz: could you create a helper functionn for this? I would call thhis function:
```
updateMappingCache( data.mappings )
```

src/lib/data.ts
```diff
@@ -37,6 +75,26 @@ const setData = (table: string, records: unknown): void => {
...
+        {}
+      )
+      mappingsDict = initialData.reduce(
+        (obj, item) => ({
+          ...obj,
+          [item.id]: item
+        }),
+        {}
+      )
+    }
```
songz: could you create a helper functionn for this? I would call thhis function:
```
setMappingCache( data.mappings )
```

src/public/client.ts
```diff
@@ -106,7 +106,7 @@ Promise.all([
   fetch('/api/statuses').then(r => r.json())
 ]).then(([mappings, statuses]: [Mapping[], Status[]]) => {
   domainList.innerHTML = ''
-  mappings
+  Object.values(mappings)
```
songz: should keep as an array.

src/api/mapping.ts
```diff
@@ -26,12 +26,12 @@ mappingRouter.post('/', async (req, res) => {
   const fullDomain = req.body.subDomain
     ? `${req.body.subDomain}.${req.body.domain}`
     : `${req.body.domain}`
-  const existingSubDomain = domainKeys.find(e => e.fullDomain === fullDomain)
+  const existingSubDomain = domainKeys[fullDomain]
```
songz: `getMappingByDomain(fullDomain)`

src/lib/data.ts
```diff
@@ -1,13 +1,17 @@
...
 
 const data: DB = {
   serviceKeys: [],
   mappings: [],
   availableDomains: []
 }
 
+let mappingsCache: MappingObj | {} = {}
+let mappingsDict: MappingObj | {} = {}
+
```
songz: Below functions, you can just call this function instead of always writing the `if` statement: `updateCache(table)`
```
const updateCache = (table) => {
  if (table === 'mappings') {
    mappingsCache = createDomainCache(data.mappings)
    mappingsDict = createIdCache(data.mappings)
  }
}
```

PR #324, opened 2019-12-08
Combining API

hwong0305:
Moved Statuses API into Mapping API.

The mapping API now pings PM2 for status of domains when fetching all domains.


src/api/mapping.ts
```diff
@@ -86,9 +86,30 @@ mappingRouter.post('/', async (req, res) => {
...
+      (statusObj, el) => ({
+        ...statusObj,
+        [el.name]: el.pm2_env.status
+      }),
+      {}
+    )
+    const fullDomainStatusMapping = domains.map(el => {
+      if (statusData[el.fullDomain]) {
+        return { ...el, status: statusData[el.fullDomain] }
+      } else {
```
songz: you don't need this `else` cuz you are returning.

Issue #288, opened 2019-11-08
Ability to download logs


songz:
**Is your feature request related to a problem? Please describe.**
When an issue is happening with my app, I would like to be able to view my server logs.

**Describe the solution you'd like**
Beside each app, I have an icon to down logs.

songz:
User 1 request
![Screen Shot 2019-12-06 at 5 04 06 PM](https://user-images.githubusercontent.com/686933/70366200-7652b600-184a-11ea-8644-973ef958384c.png)

hwong0305:
Resources for logs:

https://stackoverflow.com/questions/13106096/stream-files-in-node-express-to-client
Log files: https://stackoverflow.com/questions/55828017/what-is-the-default-location-of-pm2-log-files

PR #325, opened 2019-12-11
Logs - Implementing the ability to Download Logs

hwong0305:
#288 

- Updated code to have pm2 output logs into a specific location and filename for each app.
- implementing API to get error and output logs from the server
- Added test for API

Logs Screenshot


src/api/logs.ts
```diff
@@ -0,0 +1,29 @@
...
+import { getMappingByDomain } from '../lib/data'
+
+const logsRouter = express.Router()
+const { isProduction } = environment
+
+logsRouter.get('/err/:domain', (req, res) => {
+  const { domain } = req.params
+  const { fullDomain } = getMappingByDomain(domain)
+
+  if (!isProduction()) return res.send('OK')
```
joshgreenwell: Why not get the logs in dev? I don't have context but, a comment in the code might be nice for future devs.

joshgreenwell:
Can you add a screenshot to the PR description? I'm curious about what this feature looks like on the UI.

src/api/logs.ts
hwong0305: I updated the codebase to allow for access of logs in dev.  From the current codebase with c0d3.com, the proxy server will not run when running in dev. The user will have to navigate to the port on the server to access their apps.

hwong0305:
Updated PR description to include screenshot of the feature

src/helpers/environment.ts
```diff
@@ -5,5 +5,6 @@ export default {
   HOME: process.env.HOME || process.env.ENV || null,
   WORKPATH: process.env.WORKPATH || '/home/myproxy',
   isProduction: (): boolean =>
-    (process.env.NODE_ENV || process.env.ENV) === 'production'
+    (process.env.NODE_ENV || process.env.ENV) === 'production',
+  isTest: (): boolean => (process.env.NODE_ENV || process.env.ENV) === 'test'
```
songz: this means that during development, we have to run this with test now? I don't think devs are ready for that yet. Could we use production flag?

src/public/client.ts
```diff
@@ -41,15 +41,18 @@ class MappingItem {
     const mappingElement = document.createElement('li')
     let iconClass
     let iconColor
+    let logClass
```
songz: `let logClass = 'fa fa-file-text-o ml-1 mt-1'`

songz: why does this even have to be a variable? and putting this inside the `status` check block is kinda strange, what does status have to do with logs?

src/tests/helpers/logAdapter.ts
```diff
@@ -0,0 +1,32 @@
...
+  const options: Options = {
+    method,
+    headers: reqHeaders
+  }
+  if (body) {
+    options.body = JSON.stringify(body)
+  }
+  if (target === 'mappings') {
+    return fetch(`${apiURL}/api/mappings${path}`, options)
+  }
```
songz: remove lines 25-27

src/public/client.ts
hwong0305: This to disable UI for getting logs when pm2 doesn't return a status. (i.e for c0d3.com's application)

src/api/logs.ts
songz: @joshgreenwell `logs` are directly tied to `root pm2`, so one should be running in production flag when testing these to simulate production as much as possible.

At this point I don't see any need other than prod vs non-prod.

Usually `test` vs `stage` vs `sandbox` environment comes into importance when you start dealing with databases and billing.

PR #328, opened 2019-12-12
Unhandled rejections #313

### LINKED ISSUE #313: Unhandled rejections
**Describe the bug**
Unhandled rejections

Screen Shot 2019-12-05 at 1 10 18 PM

Wolfy64:
close #313 

If you are ok this solution I will extends it on other `Promises`

### DIFF
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -82,7 +82,7 @@ const startAppServer = (
       console.log(cyan, `myProxy is running on port ${port}!`)
       resolve(server)
     })
-  })
+  }).catch(error => console.error(`startAppServer: ${error}`))
 }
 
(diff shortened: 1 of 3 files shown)


src/server/server.ts
```diff
@@ -82,7 +82,7 @@ const startAppServer = (
       console.log(cyan, `myProxy is running on port ${port}!`)
       resolve(server)
     })
-  })
+  }).catch(error => console.error(`startAppServer: ${error}`))
```
songz: this is triggering the `new Promise` object so we should be catching here.

songz: instead, update `src/app.js` where this function is being called, and promise is being returned.

PR #329, opened 2019-12-12
fix(data race): use readFileSync on startup to avoid test failure

bryanjenningz:
Before, when you ran the tests, there was a data race because when you start up the data.ts module, there was an asynchronous readFile call and after it read the file asynchronously it would overwrite the global state. This made it so we had to overwrite the global state after calling writeFile when running our tests for our tests to work. This change makes it so you don't have to worry about that problem anymore. We now synchronously read from the file and update the state so there's no longer a data race.

### DIFF
--- a/src/lib/data.ts
+++ b/src/lib/data.ts
@@ -20,7 +20,7 @@ const updateCache = (table: string): void => {
   }
 }
 
-{
+try {
   const file = fs.readFileSync('./data.db')
   const fileData: DB = JSON.parse(file.toString() || '{}')
   data.serviceKeys = fileData.serviceKeys || []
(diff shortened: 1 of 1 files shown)


src/lib/data.ts
```diff
@@ -35,7 +30,9 @@ fs.readFile('./data.db', (err, file) => {
 
   domainToMapping = createDomainCache(data.mappings)
   idToMapping = createIdCache(data.mappings)
-})
+} catch (err) {
+  console.error(`Error reading from data.db on startup, ${err}`)
```
songz: I think `console.log` is better, because we expect this to happen when it starts.... 
```
console.log(	
      'File does not exist, but do not worry. File will be created on first save',	
      err	
    )
```

PR #330, opened 2019-12-12
 fix #315 successfully deletes the token

### LINKED ISSUE #315: Admin API Key / Secret Gen
1. UI to create API Token
2. UI to delete API Tokens
3. API to power UI (Get, create, delete)

Currently, API authentication is using ADMIN password. Highly sensitive information, we don't want to send passwords around (because it may get compromised).

Instead, we generate API tokens. These tokens can be deleted (hence invalidated) when security of the token is compromised.

All 3rd party (like c0d3.com) will be using API tokens to interact with API.

Keep existing admin password auth. Auth should support both API token and Admin password.

rkalra247:


src/api/accessToken.ts
```diff
@@ -33,4 +33,15 @@ accessTokensRouter.get('/', (req, res) => {
   res.json(tokens)
 })
 
+accessTokensRouter.delete('/:id', (req, res) => {
+  const tokens = getAccessTokens()
+  tokens.forEach((e, i) => {
```
songz: `tokens.find((e,i)=> {.....` -> This way you can have less than O(n) instead of forsure O(n)

src/tests/integration/accessToken.test.ts
```diff
@@ -45,4 +45,21 @@ describe('/api/accessTokens', () => {
...
+      name
+    })
+    expect(postResponse.status).toEqual(200)
+    const postToken = postResponse.json()
+    const tokens = await accessTokensAdapter('/', 'GET').then(r => r.json())
+    tokens.forEach((e, i) => {
+      if (e.name === name) {
+        tokens.splice(i, 1)
+      }
+    })
```
songz: you can't re implement the delete logic in the test, then you are not really testing anything.

1. POST to create an accessToken
2. Get to make sure the accessToken is there
3. Delete the accessToken
4. Get and then should receive empty (accessToken not there)

src/api/accessToken.ts
```diff
@@ -33,4 +33,16 @@ accessTokensRouter.get('/', (req, res) => {
...
+accessTokensRouter.delete('/:id', (req, res) => {
+  const tokens = getAccessTokens()
+  tokens.find((e, i) => {
+    if (e.id === req.params.id) {
+      tokens.splice(i, 1)
+      res.json(e)
+      return true
+    }
+  })
+  setData('apiTokens', tokens)
```
songz: accessTokens

PR #331, opened 2019-12-12
Refactor types

bryanjenningz:
- Added type safety to getData, setData, updateCache by using conditional types
- Fixed typo bug that came up from adding this type safety (changed 'apiTokens' to 'accessToken')
- Renamed MappingObj to MappingById
- Fixed typo in MappingObj type where there was accidentally a property on the object called 'string' instead of an object with string keys

### DIFF
--- a/src/lib/data.ts
+++ b/src/lib/data.ts
@@ -13,7 +13,7 @@ const data: DB = {
 let domainToMapping: MappingById = {}
 let idToMapping: MappingById = {}
 
-const updateCache = (table: keyof typeof data): void => {
+const updateCache = (table: keyof DB): void => {
   if (table === 'mappings') {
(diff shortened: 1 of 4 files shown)


src/lib/data.ts
```diff
@@ -10,10 +10,10 @@ const data: DB = {
   accessToken: []
 }
 
-let domainToMapping: MappingObj | {} = {}
-let idToMapping: MappingObj | {} = {}
+let domainToMapping: MappingById = {}
+let idToMapping: MappingById = {}
 
-const updateCache = (table: string): void => {
+const updateCache = (table: keyof typeof data): void => {
```
bryanjenningz: I guess this would be more simple if I changed it to just `keyof DB`

PR #335, opened 2019-12-15
[Do Not Merge] new feature(middleware for token auth)

hwong0305:
- Updated middleware for auth to have function to use accessTokens
- Fixed missing bracket typo in logs. (caught during test)
- Implement integration test for logging on with accessTokens

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -5,14 +5,15 @@ import mappingRouter from './mapping'
 import sshKeyRouter from './sshKeys'
 import accessTokensRouter from './accessToken'
 import { getAvailableDomains } from '../lib/data'
+import { setupAuth, setupTokenAuth } from '../auth'
 
 const apiRouter = express.Router()
 
-apiRouter.use('/admin', adminRouter.app)
-apiRouter.use('/logs', logsRouter)
-apiRouter.use('/mappings', mappingRouter)
-apiRouter.use('/sshKeys', sshKeyRouter)
-apiRouter.use('/accessTokens', accessTokensRouter)
+apiRouter.use('/admin', setupAuth, adminRouter.app)
+apiRouter.use('/logs', setupTokenAuth, logsRouter)
+apiRouter.use('/mappings', setupTokenAuth, mappingRouter)
+apiRouter.use('/sshKeys', setupTokenAuth, sshKeyRouter)
+apiRouter.use('/accessTokens', setupTokenAuth, accessTokensRouter)
 
 apiRouter.get('/availableDomains', (req, res) => {
   const domains = getAvailableDomains()
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -5,15 +5,15 @@ import mappingRouter from './mapping'
 import sshKeyRouter from './sshKeys'
 import accessTokensRouter from './accessToken'
 import { getAvailableDomains } from '../lib/data'
-import { setupAuth, setupTokenAuth } from '../auth'
+import { setupPwAuth, setupPwTokenAuth } from '../auth'
 
 const apiRouter = express.Router()
 
-apiRouter.use('/admin', setupAuth, adminRouter.app)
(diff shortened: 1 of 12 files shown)


src/helpers/cache.ts
```diff
@@ -1,23 +1,18 @@
...
-    }),
+    (obj, item) => ({ ...obj, [item.fullDomain]: item }),
     {}
   )
 }
 
-export { createDomainCache, createIdCache }
+const createTokenCache = (records: AccessToken[]): TokenById => {
+  return records.reduce((obj, item) => ({ ...obj, [item.id]: item }), {})
+}
```
bryanjenningz: `createTokenCache`'s implementation is the same as `createIdCache`, the only difference is their types, where `createTokenCache` takes in an array of type `AccessToken` and `createIdCache` takes in an array of type `Mapping`.

So there's a little bit of repetition which is fine because it's easy to understand what each function does and where it will be used and it's not a big deal.

I think this repetition is fine, but if we wanted to remove this repetition, we could make a generic function that does the same thing and can be applied in both situations in the same way. We could probably call it something like `mapById` since the function maps each record by the record's id property. This is how we would define this function.
```
const mapById = <T extends { id: string }>(records: T[]): { [id: string]: T | undefined } => {
  return records.reduce((obj, item) => ({ ...obj, [item.id]: item }), {})
}
```
This expression between `<` and `>` is a generic type parameter, type parameters are kind of like function parameters except they're for types, `<T>` means the type `T` can be anything, `<T extends { id: string }>` means the type `T` can be anything that has a property `id` which is type `string`, so `AccessToken` and `Mapping` would both work because they both have an `id` property that is type `string`, so we would be able to call `mapById` on an array of `AccessToken`s or an array of `Mapping`s.

This generic function is probably harder to understand so maybe just keep these functions the way they are, I just thought I would share this information because I think it's cool.

src/lib/data.ts
```diff
@@ -70,7 +84,7 @@ const getAvailableDomains = (): Domain[] => {
 }
 
 const getAccessTokens = (): AccessToken[] => {
-  const initialData = getData('accessToken') as AccessToken[] | undefined
+  const initialData = getData('accessTokens') as AccessToken[] | undefined
```
bryanjenningz: I think maybe we should make `getData` have a return type that includes `undefined` we don't need to write `as AccessToken[] | undefined`.
We can make getData like this
```
const getData = <T extends keyof DB>(table: T): DB[T] | undefined => {
```
and then we can probably remove the `as AccessToken[] | undefined` here and still have type safety

src/api/index.ts
```diff
@@ -5,14 +5,15 @@ import mappingRouter from './mapping'
...
+import { validUser, validAdmin } from '../helpers/auth'
 
 const apiRouter = express.Router()
 
-apiRouter.use('/admin', adminRouter.app)
-apiRouter.use('/logs', logsRouter)
-apiRouter.use('/mappings', mappingRouter)
-apiRouter.use('/sshKeys', sshKeyRouter)
-apiRouter.use('/accessTokens', accessTokensRouter)
+apiRouter.use('/admin', validAdmin, adminRouter.app)
```
songz: `validAdmin` should be used in the `adminRouter` as a middleware instead of here.

src/api/index.ts
```diff
@@ -5,14 +5,15 @@ import mappingRouter from './mapping'
...
+apiRouter.use('/logs', validUser, logsRouter)
+apiRouter.use('/mappings', validUser, mappingRouter)
+apiRouter.use('/sshKeys', validUser, sshKeyRouter)
+apiRouter.use('/accessTokens', validUser, accessTokensRouter)
```
songz: maybe just put this in front:
```
apiRouter.use(validUser)
apiRouter.use('/admin'...)
apRouter.use('/logs'....)
apiRouter.use('/mappings'....)
....
```

hwong0305:
This PR has been restructured into separate PRs. Closing PR.

PR #341, opened 2019-12-19
2/3 - Added logic for accessToken Middleware and test.

hwong0305:
This is a 3 part commit for implementing accessToken middleware for MyProxy

**PR:1** - ~Refactoring AccessToken and add Caching~
**PR:2** - Added logic for accessToken Middleware and test.
**PR:3** - Updating repo to implement logic for accessToken Middleware.

### DIFF
--- a/src/auth/index.ts
+++ b/src/auth/index.ts
@@ -40,6 +40,7 @@ const setupAccessToken = (req, res, next): void => {
   }
   if (adminPass === hashPass(pass)) {
     req.user.isAdmin = true
+    req.user.isUser = true
     return next()
   }
   if (authorization || access) {
--- a/src/auth/index.ts
+++ b/src/auth/index.ts
@@ -34,7 +34,6 @@ const setupAuth = password => {
 }
 
 // This will be changed into setupAuth at a PR later
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const setupAccessToken = (req, res, next): void => {
   const { adminPass } = req.cookies
   const { access = '', authorization = '' } = req.headers
@@ -55,7 +54,7 @@ const setupAccessToken = (req, res, next): void => {
     req.admin = true
     return next()
   }
-  return res.status(401).send('Unauthorized')
+  return next()
 }
 
-export { setupAuth, setPass, isCorrectCredentials }
+export { setupAuth, setPass, setupAccessToken, isCorrectCredentials }
--- a/src/auth/index.ts
+++ b/src/auth/index.ts
@@ -33,25 +33,17 @@ const setupAuth = password => {
 // This will be changed into setupAuth at a PR later
 const setupAccessToken = (req, res, next): void => {
   const { adminPass } = req.cookies
-  const { access = '', authorization = '' } = req.headers
-  req.user = {
-    isAdmin: false,
-    isUser: false
-  }
-  if (adminPass === hashPass(pass)) {
-    req.user.isAdmin = true
-    req.user.isUser = true
+  const { authorization = '' } = req.headers
+  if (
+    adminPass === hashPass(pass) ||
+    isCorrectCredentials(authorization, pass)
+  ) {
+    req.user = { isAdmin: true, isUser: true }
     return next()
   }
-  if (authorization || access) {
-    const correctPw = isCorrectCredentials(authorization, pass)
-    const correctToken = isValidAccessToken(access)
-    if (!correctPw && !correctToken) return res.status(401).send('Unauthorized')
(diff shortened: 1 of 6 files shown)


hwong0305:
The validUI middleware will be used to redirect users to the login page if they don't have a password or cookie. 

I did not want this logic in the API because it would be confusing to users when their API returns 200 but instead of returning a JSON of all the information, they get a html page with a response of 200.  This is hard to debug for developers using our API.

src/auth/index.ts
```diff
@@ -21,4 +33,28 @@ const setupAuth = password => {
...
+    }
+    if (correctToken) {
+      req.user = true
+    }
+    return next()
+  }
+  if (adminPass === hashPass(pass)) {
+    req.admin = true
+    return next()
+  }
```
songz: I would have this up, to before the `if (authorization || access)` line

src/auth/index.ts
```diff
@@ -21,4 +33,28 @@ const setupAuth = password => {
   }
 }
 
-export { setupAuth, isCorrectCredentials }
+// This will be changed into setupAuth at a PR later
+const setupAccessToken = (req, res, next): void => {
+  const { adminPass } = req.cookies
+  const { access = '', authorization = '' } = req.headers
+
```
songz: Since an admin is also a user, the standard practice is to make user object:
```
req.user = {
  isAdmin: true/false,
  isLoggedIn: true/false // This is to prevent future accidents when other devs accidentally set req.user = 'undefined' or something
}
```

src/helpers/authentication.ts
```diff
@@ -0,0 +1,16 @@
+const validUIAccess = (req, res, next): void => {
+  if (!req.admin && !req.user) return res.redirect('/login')
```
songz: Continuing from the previous comment,
```
if ( (req.user || {}).isValid )....
```

src/helpers/authentication.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const validAdmin = (req, res, next): void => {
+  if (!req.admin) return res.status(401).send('Unauthorized')
+  return next()
+}
+
+const validUser = (req, res, next): void => {
+  if (!req.admin && !req.user) return res.status(401).send('Unauthorized')
+  return next()
+}
```
songz: I don't think we would need these 2 hhelper funnctions.

songz: I don't think you need these 2 functions because our code is structured in such a way these fucntions shhouldn't be used more than once. For example, you could simply put `validAdmin` inside the admin api router. and `validUser` inside the main api router. 

Main API router -> checks if user is valid, rrejects if not.
Admin API router -> (at this point, we know user exist). Check for the user's admin status.

src/auth/index.ts
```diff
@@ -1,13 +1,25 @@
...
+import { getTokenById } from '../lib/data'
+let pass = ''
 
 const isCorrectCredentials = (password: string, correct: string): boolean => {
   const adminPassword = hashPass(password)
   const userPassword = hashPass(correct)
   return userPassword === adminPassword
 }
 
+const isCorrectAccessToken = (token: string): boolean => {
```
songz: `isValidAccessToken`. `correct` (implies 1) could be confusing

src/auth/index.ts
```diff
@@ -1,13 +1,25 @@
...
+  if (getTokenById(token)) return true
+  return false
```
songz: not importantn, but `return !!getTokenById(token)` would work too.

src/auth/index.ts
```diff
@@ -21,4 +30,31 @@ const setupAuth = password => {
...
+// This will be changed into setupAuth at a PR later
+const setupAccessToken = (req, res, next): void => {
+  const { adminPass } = req.cookies
+  const { access = '', authorization = '' } = req.headers
+  req.user = {
+    isAdmin: false,
+    isUser: false
+  }
+  if (adminPass === hashPass(pass)) {
+    req.user.isAdmin = true
```
songz: admins should be users too, because of the `validUi` check below: `if (!req.user.isUser) return res.redirect('/login')`

this code block should be above, right below: `  const { adminPass } = req.cookies`

src/helpers/authentication.ts
```diff
@@ -0,0 +1,6 @@
+const validUIAccess = (req, res, next): void => {
+  if (!req.user) res.redirect('/login')
```
songz: `return res.redirect`

without return, you may be trying to send back multiple responses for one request, resulting in warnings.

PR #343, opened 2019-12-20
Sample PR for provider contribution 1/ Adding name.com

Wolfy64:
Sometimes to clone it I have to replace the hostname by the IP address...
Screen Shot 2019-12-20 at 3 02 55 AM


src/api/admin.ts
```diff
@@ -45,6 +45,7 @@ app.post('/sslCerts', async (req, res) => {
     setData('availableDomains', domains)
     return res.json(serviceResponse)
   } catch (err) {
+    console.error('POST /sslCerts err:', err)
```
songz: THANNK YOU.

src/api/admin.ts
```diff
@@ -158,7 +159,10 @@ app.patch('/providerKeys/:id', (req, res) => {
 })
 
 app.get('/providers', async (_, res) => {
-  const data = await Promise.all([providers.dns_gd.getDomains()])
+  const data = await Promise.all([
+    providers.dns_gd.getDomains(),
+    providers.dns_namecom.getDomains()
```
songz: I would prefer to have new providers not hhave to add any code in here. Could we change the logic here to use a `filter` and then `Promise.all`? This way, new providers don't have to update this when the want thheir service to be supported.

Wolfy64: Sure I was thinking about to do a generic function that look at `src/api/serviceConfig.ts` or even maybe a config file with a list of all providers that work with `acme.sh` then will we just have to uncomment provider to enable it.
```
[
  dns_gd,
  dns_namecom,
  // dns_provider,
 //  ... and so on
]
```

src/providers/index.ts
```diff
@@ -1,5 +1,6 @@
 import { ServiceConfig } from '../types/general'
 
 export default {
-  dns_gd: require('./goDaddy')
+  dns_gd: require('./goDaddy'),
+  dns_namecom: require('./nameDotCom')
```
songz: we can do a `fs.readdir` here and make thhis generic so that new api providers don't have to set this up.

src/api/admin.ts
songz: or this:

```
const requests = Object.keys(providers).filter((name) => {
  return providers[name].getDomains
}).map( (name) => {
  return providers[name].getDomains()
})

const data = await Promise.all( requests )
```

PR #344, opened 2019-12-20
fix #340 add nav bar with AccessTokens link

### LINKED ISSUE #340: no way to get ito AccessTokens page
There's no way to navigate into or out of accesstokens page

rkalra247:


views/admin/accessTokens.ejs
```diff
@@ -2,8 +2,35 @@
...
+   <div class="flexContainer" style="display: flex;">
     <h1>AccessTokens</h1>
 
+    <div class="navBar" style="margin-left: 513px">
+      <a 
+        class="btn-outline-success btn"
+        style="margin-right: auto;"
+        href="/sshKeys"
+        >
+        Managge SSH KEYS
```
songz: `Manage`

views/admin/accessTokens.ejs
```diff
@@ -2,8 +2,35 @@
...
+        class="btn-outline-primary btn"
+        style="margin-right: auto;"
+        href="/"
+        >
+        My Proxy
+      </a>
+      <a 
+        class="btn-outline-primary btn"
+        style="margin-right: auto;"
+        href="/sshKeys"
```
hwong0305: /admin

views/admin/accessTokens.ejs
```diff
@@ -16,7 +43,7 @@
     <a 
       class="btn-outline-success btn createToken"
       style="margin-right: 100px;"
-      href="/api/accessTokens"
+      href="/admin/accessTokens"
```
hwong0305: This should be a button. It doesn't need a `href` element since it already has logic to create the button on the server.

songz: yeah.... we'll take this up later.

PR #346, opened 2019-12-21
Refactor Providers to prepare for Sample PR

Wolfy64:
It's a way much cleaner to have only one place to add/change providers values.


Wolfy64:
Into `src/types/general.ts`
```js
type ServiceConfig = {
  dns_gd: ProviderService
}
```
How can I change this type to be more generic?
It would be better if a user don't have to add a line each time we add a new provider
```js
type ServiceConfig = {
  dns_api: ProviderService // That would be ok?
}
```

src/api/serviceConfig.ts
```diff
@@ -1,7 +1,9 @@
+import { GoDaddy } from '../constants/providers'
+
 const serviceConfig: unknown = {
-  dns_gd: {
-    name: 'Godaddy',
-    keys: ['GD_Key', 'GD_Secret']
+  [GoDaddy.DNS_API]: {
+    name: GoDaddy.NAME,
+    keys: [GoDaddy.PRIMARY_KEY, GoDaddy.SECONDARY_KEY]
```
songz: I don't see this change having any benefits, its actually makes things more confusing right?

Wolfy64: Really ?
The benefit is to have only one place to set up new provider
At `src/constants/providers.ts`
```js
export const GoDaddy = {
  NAME: 'GoDaddy',
  DNS_API: 'dns_gd',
  PRIMARY_KEY: 'GD_Key',
  SECONDARY_KEY: 'GD_Secret',
  SERVICE: 'https://api.godaddy.com'
} as ProviderInfo
```
Maybe  I can improve it with a function that will read `src/constants/providers.ts` and adding them within `src/api/serviceConfig.ts` ?

Wolfy64:
**One file to rule them all 🙂**
I refactored this PR so now we fallow acme.sh naming convention.
To add a new provider the contributor will have to edit only 2 files:
- Adding the provider info to the `providerList` within `./src/providers/index.ts`
- Create the new provider file with it own logic within `./src/providers`

PR #354, opened 2019-12-24
Change minimum port for setting up myProxy App to 3000 instead of 3001.

hwong0305:


songz:
not necessary anymore.

PR #356, opened 2019-12-25
Add doc for Name.com

Wolfy64:


README.md
```diff
@@ -138,10 +138,145 @@ The following steps will guide you through how to setup your development environ
...
++  }
+] as ProviderInfo[]
+```
+
+### Create the Typescript file
+
+Now we will add the `namecom.ts` within `src/providers/namecom.ts`
+
+Please you can rely on existing providers the logic will be the same.
+Quick eplanation of the code.
```
hwong0305: explanation

README.md
```diff
@@ -138,10 +138,145 @@ The following steps will guide you through how to setup your development environ
...
+const { name, dns, keys, service } = provider
+```
+
+Will find and setup the provider keys `Namecom_Username` and `Namecom_Token`
+
+```js
+const getKeys = () => {...} ServiceKey[]
+```
+
+This method is **mandatory** and will return your domains
```
hwong0305: I think we can take this line out. Everything in the doc should be mandatory.

README.md
```diff
@@ -138,10 +138,145 @@ The following steps will guide you through how to setup your development environ
...
+
+```js
+export const getDomains = () => {...} Promise<Provider>
+```
+
+This method is **mandatory** and will setup your hostname record
```
hwong0305: Same comment as above.

PR #361, opened 2020-01-07
Fix: subdomain names with upper case letters does not work properly

hwong0305:
- Update codebase to convert subdomain and domain names to lowercase.
- Update test to reflect these changes.

Added a migration plan for existing database. Users can run `cp ./data.db ./data.db.backup` to back up their database. Then `npm run migration` to update their exisiting domains. However, users will not be able to view the status of the domain unless they delete and recreate the app.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "autofix": "npm run prettier:write && npm run eslint:fix",
     "server": "./scripts/setup.sh; pm2 startOrRestart ./scripts/prod.config.js --env production --update-env",
     "server:stop": "pm2 stop myProxy-prod",
-    "migration": "node ./dataMigration.js"
+    "migrate": "node ./dataMigration.js"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -18,7 +18,8 @@
     "lint": "npm run prettier:check && npm run eslint:check",
     "autofix": "npm run prettier:write && npm run eslint:fix",
     "server": "./scripts/setup.sh; pm2 startOrRestart ./scripts/prod.config.js --env production --update-env",
-    "server:stop": "pm2 stop myProxy-prod"
+    "server:stop": "pm2 stop myProxy-prod",
+    "migration": "node ./dataMigration.js"
```
songz: `migrate` -> use a verb

hwong0305:
> You should also address routing.
> 
> 1. Sally creates a `HelloCooking.c0d3.com` domain. In her mind, she typed `HelloCooking.c0d3.com`. That's the website she gave to all her friends.
> 2. Since we changed everything to lowercase without user knowing, what happens when Sally's site visits `HelloCooking.c0d3.com`? Could you make sure it doesn't say `Not Found`?

Domain names aren't case sensitive. If Sally tells her friends to go to `HelloCooking.c0d3.com`. When her friends put it into the browser will always send the request to `hellocooking.c0d3.com`.

Issue #367, opened 2020-03-18
Admin key leak


songz:
**Describe the bug**
Admin key is present in `deploy.config.js` file when you download it.

hwong0305:
closed by #371

PR #371, opened 2020-05-16
Fix Issue #367 - Admin Key Leak

hwong0305:


src/api/mapping.ts
```diff
@@ -48,6 +48,7 @@ mappingRouter.post('/', async (req, res) => {
   prodConfigApp.script = `npm run start:myproxy << /home/myproxy/.pm2/logs/${fullDomain}-out.log`
   prodConfigApp.error_file = `/home/myproxy/.pm2/logs/${fullDomain}-err.log`
   prodConfigApp.merge_logs = true
+  delete prodConfigApp.env_production.ADMIN
```
hwong0305, line 51: `prodConfigApp` is being imported from `prod.config.js` which is used to run myProxy.

The simplest solution was to just delete the value from the object.

PR #374, opened 2020-08-28
Block SSH shell access to git repositories

ggwadera:
This pull request contains changes to block shell access to the server through the SSH link used as the git repository for domains.
The access is blocked by setting up a `git` user account restricted to only Git-related activities through a limited shell tool called `git-shell` that comes with Git. [Reference](https://git-scm.com/book/en/v2/Git-on-the-Server-Setting-Up-the-Server).

What this PR does:

- Modify the `setup.sh` script to create the `git` user account in the `myproxy` group, and set the correct file permissions.
  - `git` user has restricted shell access through `git-shell`
  - adds sudoers rule so the `git` user can run `sudo -u myproxy pm2` without password
  - disable server MOTD for the `git`user so it doesn't print when someone tries to SSH into the server.
  - adds a script to print a message when someone tries to SSH into the server.

- Modify the `post-receive` script so the app is started by the `myproxy` user instead of `git`.

- Change the `exec` commands in `/src/api/mappings.ts` options to use the correct group id. (Added new helper function for this)

- Change the `authorized_keys` function to save the SSH keys under the `git` user and with additional options.

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -15,14 +15,27 @@ if [ ! -d "./acme.sh" ] ; then
   cd ../
 fi
 if [ ! -d "/home/myproxy" ] ; then
+  # Add users
   sudo useradd -m -c "myproxy" myproxy -s /bin/bash -p $(echo $ADMIN | openssl passwd -1 -stdin) -d "/home/myproxy"
+  sudo useradd -m -G myproxy -s $(which git-shell) -p $(echo $ADMIN | openssl passwd -1 -stdin) git
+  # Add sudoers rule for git user to run pm2 as myproxy, without password
+  echo "git ALL = (myproxy) NOPASSWD: /usr/bin/pm2" > /etc/sudoers.d/git
+  # Create folders
   mkdir /home/myproxy/.ssh
+  mkdir /home/git/.ssh
   mkdir /home/myproxy/.scripts
+  # Copy ssh keys and scripts
   cp ~/.ssh/authorized_keys /home/myproxy/.ssh/authorized_keys
+  cp ~/.ssh/authorized_keys /home/git/.ssh/authorized_keys
   cp ./scripts/post-receive /home/myproxy/.scripts/post-receive
   cp ./scripts/pre-receive /home/myproxy/.scripts/pre-receive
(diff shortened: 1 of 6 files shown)


scripts/setup.sh
```diff
@@ -30,6 +30,8 @@ if [ ! -d "/home/myproxy" ] ; then
   cp ./scripts/post-receive /home/myproxy/.scripts/post-receive
   cp ./scripts/pre-receive /home/myproxy/.scripts/pre-receive
   cp ./scripts/gitignore /home/myproxy/.scripts/.gitignore
+  # Disable SSH MOTD message for git user
+  touch /home/git/.hushlogin
```
ggwadera, line 34: This disables the MOTD sent from the server when someone tries to SSH in. For example:

```bash
$ ssh git@challenges.ggwadera.xyz
PTY allocation request failed on channel 0
Welcome to Ubuntu 20.04.1 LTS (GNU/Linux 5.4.0-42-generic x86_64)

 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/advantage

  System information as of Tue Sep  1 13:15:29 UTC 2020

  System load:  0.72               Users logged in:          1
  Usage of /:   18.9% of 24.06GB   IPv4 address for docker0: 172.17.0.1
  Memory usage: 48%                IPv4 address for eth0:    157.245.253.104
  Swap usage:   0%                 IPv4 address for eth0:    10.10.0.5
  Processes:    130

13 updates can be installed immediately.
0 of these updates are security updates.
To see these additional updates run: apt list --upgradable


fatal: Interactive git shell is not enabled.
hint: ~/git-shell-commands should exist and have read and execute access.
Connection to challenges.ggwadera.xyz closed.
```

Becomes: 

```bash
$ ssh git@challenges.ggwadera.xyz
PTY allocation request failed on channel 0
fatal: Interactive git shell is not enabled.
hint: ~/git-shell-commands should exist and have read and execute access.
Connection to challenges.ggwadera.xyz closed.
```

songz:
everything worked out for me, running https://huevon.club/ with new server. Cannot ssh in (good security)

Screen Shot 2020-09-01 at 12 38 01 AM

PR #378, opened 2020-09-02
[FIX] Disabling force pushing to remote repository

hwong0305:
# Problem Statement
There is a security flaw with `deploy.config.js` that allows the user to execute any arbitrary script with pm2.. The current git configuration allows user to overwrite what's already on the server with a force push.

# Solution
Change git configuration to reject the pull request if it can't fast forward.  This will prevent users from overwriting `deploy.config.js` without downloading it first. Since `deploy.config.js` has been added to `.gitignore`, it will prevent users from modifying `deploy.config.js`

# Notes

The `pre-receive` hook has to be modify to delete the `prod` branch if user fails to git push. Otherwise, `pre-receive` will error out forcing the user to delete the repo and push again instead of adding new commits.


songz (approved):

songz:
no longer an issue.

PR #379, opened 2020-09-04
Run apps inside Docker containers

### LINKED ISSUE #281: Ability to turn app on and off
**Is your feature request related to a problem? Please describe.**
My computer encounters high memory usage with too many apps.

**Describe the solution you'd like**
Control buttons (maybe play/pause button?) to allow me to control to stop and start an app.

**Additional context**
API to start/stop an app, that executes a `pm2` command.

### LINKED ISSUE #377: Run apps inside Docker containers
**Is your feature request related to a problem? Please describe.**
Running the apps without proper isolation could lead to security issues and data leaks. Currently, the apps are managed using `pm2` and they are executed without any isolation.

**Describe the solution you'd like**
Run each domain app in a separate Docker container, achieving file system and network isolation.

This can be done by running Docker commands on a shell using `exec` or with a library like [dockerode](https://github.com/apocas/dockerode/) which uses the Docker API.

What's needed:
- [ ] Install and setup docker in `setup.sh` (is this needed?).
- [x] Start the apps using a Docker container instead of `pm2` when the user pushes a new commit.
    - [x] Bind the correct port for the domain to the container.
    - [x] Pass the domain port as a environment variable to the container.
    - [x] Bind the domain directory to a folder inside the container.
    - [x] Start the app inside the container (running with `node .`, needs to set `main` inside `package.json`).
- [x] Get the logs (stdout and stderr) from the container (redirect to a file or get through Docker API when requested?).
- [x] Get existing containers list and status (running or stopped).
- [x] Stop and delete the container when domain is deleted.

**Describe alternatives you've considered**
Running each app as a different user with specific permissions only for the domain directory.
- this could lead towards more confusion involving user and permissions management in the server.
- this would start a new `pm2` daemon for each user, which is not efficient.

### References

[Docker Installation Guide on Ubuntu](https://docs.docker.com/engine/install/ubuntu/)
[Docker Engine API Reference](https://docs.docker.com/engine/api/latest/)
[Docker and Node.js Best Practices](https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md)

ggwadera:
Closes #377  
Closes #281

This PR will change the apps process management from `pm2` to `docker`, running the apps inside containers to achieve better isolation, this way protecting the main system from malicious apps.

Now docker is a requirement to run myProxy.

There's been a lot of changes, so read and review this PR carefully. I'm always open to your suggestions or questions.

### PR Summary

- [x] Create a new container when a new domain is created.
    - [x] Bind the correct port for the domain to the container.
    - [x] Pass the app port as a environment variable to the container.
    - [x] Bind the domain directory to a folder inside the container.
    - [x] Set the container default command to start the app.
- [x] Start the apps using a Docker container instead of `pm2` when the user pushes a new commit.
- [x] Get the container logs (stdout and stderr).
- [x] Get existing containers list and status (running or stopped).
- [x] Stop and delete the container when domain is deleted.
- [x] Start/Restart/Stop containers through the client

## Changed

- `setup.sh` script
    - checks if `docker` is installed before executing, exiting if it's not
    - added more checks so the script does not stop halfway and does not executes unnecessary commands
    - pulls the necessary node docker image (`node:alpine` ([Alpine](https://alpinelinux.org/) is a lightweight linux distro, the image is based on that))

- `post-receive` script
    - starts the app container by sending a request through `curl` to the docker socket on the machine

- `api/mappings.ts`
    - replaced every use of `pm2` to use `docker` instead, using it's API instead of exec commands
    - when a domain is created, a new container is created with all the needed configurations like port and directory bindings
    - the domain id is set to be the same as the container id
    - on a `git push` the container is started through the `post-receive`script
    - the delete feature was changed to also remove the container
    - the apps are started  by running `node .` in the app directory. For this to work, `package.json` needs to be configured correctly by setting the app entry point file in the `main` property (for example `"main": "index.js"`).

- `api/logs.ts`
    - merged the `stdout` and `stderr` endpoints into a single one
    - now returns the logs from docker

- `public/client.ts`
    - changed the logs buttons to work with the new parameters
    - changed the domain status check to the docker equivalent

## Added

- new package dependency: [`dockerode`](https://github.com/apocas/dockerode/), used to communicate with docker through it's API

- new docker helper functions in `helpers/docker.ts`

- new query params for the logs api
    - `follow: boolean`: works like `heroku logs tail` where the log keeps updating with new changes. Does not work in the browser. Useful in the terminal when used with `curl`.
    - `tail: number`: returns the last n lines from the log.

- new start and stop buttons in the client webpage, with the respective endpoints in the mappings api
    - start/restart: `/api/mappings/{id}/start`
    - stop: `/api/mappings/{id}/stop`

image

## Removed

- clear logs feature
    - docker doesn't provide an easy way to clear it's logs, so I though better to remove this
    - as an alternative, we can get the log's tail (last n lines) by a query param in the url: `/api/logs/{stream}/{domain}?tail={numLines}`

- `deploy.config.js` is not needed anymore, all the domain configs are set when the container is created

## Possible features to be developed in the future

1. Add tests for the docker integration (would this require the dev/test env to have docker installed in their system?)
2. Integrate the c0d3-cli tool to deploy domains through the api

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -96,4 +96,11 @@ if [ ! -f "./data.db" ] ; then
 fi
 
 # pull node docker image
-docker pull node:alpine
+if docker ps > /dev/null 2>&1; then
+  docker pull node:alpine
+else
+  echo "WARNING: Couldn't run docker commands"
+  echo "WARNING: Make sure your user has the right permissions"
+  echo "WARNING: Go to this link to setup docker to run without root"
+  echo "WARNING: https://docs.docker.com/engine/install/linux-postinstall/"
+fi
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+# Check if docker is installed and stop the script if it's not
+if ! command docker &> /dev/null; then
+  echo "myProxy requires Docker to run"
+  echo "Docker installation instructions: https://docs.docker.com/engine/install/"
+  exit
+fi
+
 if ! command node -v &>/dev/null; then
   sudo apt-get install curl
   curl -sL https://deb.nodesource.com/setup_12.x | sudo -E bash -
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -18,8 +18,11 @@ if [ ! -d "/home/myproxy" ] ; then
   # Add users
   sudo useradd -m -c "myproxy" myproxy -s /bin/bash -p $(echo $ADMIN | openssl passwd -1 -stdin) -d "/home/myproxy"
   sudo useradd -m -G myproxy -s $(which git-shell) -p $(echo $ADMIN | openssl passwd -1 -stdin) git
-  # Add sudoers rule for git user to run pm2 as myproxy, without password
-  echo "git ALL = (myproxy) NOPASSWD: /usr/bin/pm2" > /etc/sudoers.d/git
+  # Add myproxy and git to the docker group so they can run the commands
+  sudo groupadd docker
+  sudo usermod -aG docker myproxy
+  sudo usermod -aG docker git
+  newgrp docker
   # Create folders
   mkdir /home/myproxy/.ssh
   mkdir /home/git/.ssh
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -44,30 +44,21 @@ if ! user_exists myproxy; then
   echo "Creating user: myproxy"
   sudo useradd -m -c "myproxy" myproxy -s /bin/bash -p $(echo $ADMIN | openssl passwd -1 -stdin) -d "/home/myproxy"
   sudo usermod -aG docker myproxy
+  mkdir -p /home/myproxy/.ssh
+  mkdir -p /home/myproxy/.scripts
 fi
 
 if ! user_exists git; then
   echo "Creating user: git"
   sudo useradd -m -G myproxy -s $(which git-shell) -p $(echo $ADMIN | openssl passwd -1 -stdin) git
   sudo usermod -aG docker git
+  mkdir -p /home/git/.ssh
   # Disable SSH MOTD message for git user
   touch /home/git/.hushlogin
   # Add git-shell message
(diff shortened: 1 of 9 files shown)


scripts/setup.sh
```diff
@@ -1,49 +1,99 @@
...
   cp ./scripts/no-interactive-login /home/git/git-shell-commands/no-interactive-login
   chmod +x /home/git/git-shell-commands/no-interactive-login
-  # fix file permissions
-  chown myproxy:myproxy -R /home/myproxy/
-  chown git:git -R /home/git/
-  chmod 2775 -R /home/myproxy/
+fi
+
+if [ ! -d "/home/myproxy/.ssh" ]; then 
+  mkdir /home/myproxy/.ssh
```
songz, line 62: I think there's a flag `-p` to make if doesn't exist.

songz: `mkdir -p /home/myproxy/.ssh`

scripts/setup.sh
```diff
@@ -1,49 +1,99 @@
...
 fi
+
+cp ./scripts/post-receive /home/myproxy/.scripts/post-receive
+cp ./scripts/pre-receive /home/myproxy/.scripts/pre-receive
+cp ./scripts/gitignore /home/myproxy/.scripts/.gitignore
+
+# fix file permissions
+chown myproxy:myproxy -R /home/myproxy/
+chown git:git -R /home/git/
+chmod 2775 -R /home/myproxy/
```
songz, line 90: could you add a comment on `2775` and what it means?

scripts/setup.sh
ggwadera: Removed the checks and add the `-p` to the `mkdir` commands

scripts/setup.sh
ggwadera: Comments added!

songz (approved):

songz:
approved - to onboard new user, did not review thoroughly.

Nice comments in the code, at a high level everything makes sense.

greenc123 (approved):

PR #383, opened 2020-09-09
Manage environment variables

### LINKED ISSUE #382: [Feature] Set environment variables for an app
Currently there is no way for an user to set environment variables for their apps if they need to. This feature would enable a new configuration option for the apps, so an user can add, edit, or remove environment variables for an app.

### Goals

* Add new button in the client frontend that an user can click and then add, edit, or remove environment variables for their app.
* Handle the environment variables modifications through a new API endpoint in the server.

### Design

The environment variables are stored inside the docker container configuration files, and we can get those variables by inspecting a container with the docker API, so there's no need to store them somewhere else.

To modify the environment variables for a existing container, we have to recreate the container with the new variables.

In summary, the steps for this feature are:

1. User presses the `Manage` button for the domain the user wants to edit.

image

2. The user is redirected to a new page (`/manage/${fullDomain}`) where the user can manage the environment variables (add, modify, delete).

image

* The `Add` button creates a new input form at the end of the list.
* The `Remove` button will remove that environment variable from the list.
* The `Submit` button will send a POST request to the API with the environment variables on the list.
* The `NAME` input is automatically converted to uppercase and validated to contain only letters, numbers and/or underscore.

image

3. The client queries the server for the container existing environment variables.

    1. Server queries docker to inspect the container and get the container info back.

    2. Server filters out the default environment variables like `NODE_ENV` and `PORT`, then returns the filtered variables.

    * The app environment variables are stored inside the container config file generated by Docker [(API reference)](https://docs.docker.com/engine/api/v1.40/#operation/ContainerInspect). For example:

    ```js
    "Env": [
      "NODE_ENV=production",
      "PORT=3000",
      "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
      "NODE_VERSION=14.9.0",
      "YARN_VERSION=1.22.5"
    ],
    ```

4. User presses the submit button to send the new environment variables to the server.

5. If the container is running, warns the user to confirm that the app will be restarted and asks if it's OK.

6. The server stops and removes the existing container.

7. The server recreates the container with the new environment variables and starts the container.

    * Example Docker configuration with a new environment variable

    ```js
    "Env": [
      "NODE_ENV=production",
      "PORT=3000",
      "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
      "NODE_VERSION=14.9.0",
      "YARN_VERSION=1.22.5",
      "MY_EXAMPLE_VAR=my_example_var_value"
    ],
    ```

#### GET: `/api/mappings/${fullDomain}/environment`

This endpoint will query the Docker API to inspect the container configurations, extract the current environment variables, and then filter out the default variables (`NODE_ENV`,  `PORT`,  `PATH`, `NODE_VERSION`, `YARN_VERSION`), returning the non-default variables with name and value.

Response sample (success):

```js
{
  variables: {
    NAME1: "value1",
    NAME2: "value2"
  }
}
```

Response sample (failure):

```js
{
  message: "error message"
}
```

#### PUT: `/api/mappings/${fullDomain}/environment`

This will save the new environment variables for the app. To do this, the container needs to be recreated with the new environment variables appended in the container configuration options (like restarting an app to pass a new env variable). After recreating, the container is started.

As we are using the same ID from the Docker container as the domain mapping ID, this will also update the mapping ID to the new container ID and update the `data.db` file.

Request body sample:
```js
{
  variables: {
    NAME1: "value1",
    NAME2: "value2"
  }
}
```

Response sample (success): status code 204, no body.

Response sample (failure):

```js
{
  message: "error message"
}
```

ggwadera:
Closes #382 

This PR will add a new feature to MyProxy, enabling the users to manage the environment variables for their domains.

See the linked issue for the design documentation and UI previews.

### Added

* New `Manage Domain` page (`/manage/${fullDomain}`), where the environment variables for a domain can be configured.

* New API endpoints to handle environment variables management for a domain.
    * GET: `/api/mappings/${fullDomain}/environment`: fetches the current env for the domain
    * PUT: `/api/mappings/${fullDomain}/environment`: configures the new env for the domain, by recreating the container.

* `inspectContainer` helper function in `docker.ts`, to query the docker API for container information, used to get the current environment variables.

* `updateDomain` helper function in `data.ts`, to save the updated mapping data to the file.

### Changed

* `createContainer` helper function in `docker.ts` to receive optional environment variables array to include in the container creation.

* Updated bootstrap version from the CDN, as some of the classes used in the new page weren't working (`text-monospace` for example).

### Screenshots

image

image

### DIFF
--- a/src/helpers/docker.ts
+++ b/src/helpers/docker.ts
@@ -5,6 +5,14 @@ import environment from '../helpers/environment'
 
 const docker = new Docker({ socketPath: '/var/run/docker.sock' })
 
+type DockerError = {
+  reason: string
+  statusCode: number
+  json: {
+    message: string
+  }
+}
+
 const getContainersList = async (): Promise<Docker.ContainerInfo[]> => {
   const containers = await docker.listContainers({ all: true })
   return containers
@@ -29,7 +37,8 @@ const getContainerLogs = async (
 
 const createContainer = async (
   fullDomain: string,
-  port: number
+  port: number,
+  environmentVariables: string[] = []
 ): Promise<string> => {
   const workPath = path.resolve(environment.WORKPATH, fullDomain)
   return docker
@@ -42,7 +51,7 @@ const createContainer = async (
       },
       Tty: false,
       WorkingDir: '/home/node/app',
-      Env: ['NODE_ENV=production', 'PORT=3000'],
+      Env: ['NODE_ENV=production', 'PORT=3000', ...environmentVariables],
       HostConfig: {
         Binds: [`${workPath}:/home/node/app`],
         RestartPolicy: {
@@ -86,11 +95,18 @@ const removeContainer = async (id: string): Promise<unknown> => {
   return container.remove({ v: true, force: true })
 }
 
+const inspectContainer = (id: string): Promise<Docker.ContainerInspectInfo> => {
+  const container = docker.getContainer(id)
+  return container.inspect()
+}
+
 export {
   getContainersList,
   getContainerLogs,
   createContainer,
   startContainer,
   stopContainer,
-  removeContainer
+  removeContainer,
+  inspectContainer,
+  DockerError
 }
--- a/src/helpers/docker.ts
+++ b/src/helpers/docker.ts
@@ -5,14 +5,6 @@ import environment from '../helpers/environment'
 
 const docker = new Docker({ socketPath: '/var/run/docker.sock' })
 
-type DockerError = {
-  reason: string
-  statusCode: number
-  json: {
-    message: string
-  }
-}
-
 const getContainersList = async (): Promise<Docker.ContainerInfo[]> => {
   const containers = await docker.listContainers({ all: true })
   return containers
@@ -107,6 +99,5 @@ export {
   startContainer,
   stopContainer,
   removeContainer,
-  inspectContainer,
-  DockerError
+  inspectContainer
 }
--- a/src/public/client.ts
+++ b/src/public/client.ts
@@ -95,6 +95,13 @@ class MappingItem {
           ${data.gitLink}
         </small>
       </div>
+      <a
+        href="/manage/${data.fullDomain}"
+        class="btn btn-sm btn-outline-primary mr-3 manageButton"
+        role="button"
+      >
+        Manage
+      </a>
       <button
         class="btn btn-sm btn-outline-success mr-3 startButton"
         type="button"
--- a/src/public/client.ts
+++ b/src/public/client.ts
@@ -97,32 +97,32 @@ class MappingItem {
       </div>
       <a
         href="/manage/${data.fullDomain}"
-        class="btn btn-sm btn-outline-primary mr-3 manageButton"
+        class="btn btn-sm btn-outline-primary mr-3 manage-button"
         role="button"
       >
         Manage
       </a>
       <button
-        class="btn btn-sm btn-outline-success mr-3 startButton"
+        class="btn btn-sm btn-outline-success mr-3 start-button"
         type="button"
       >
         Start/Restart
       </button>
       <button
-        class="btn btn-sm btn-outline-warning mr-3 stopButton"
+        class="btn btn-sm btn-outline-warning mr-3 stop-button"
         type="button"
       >
         Stop
       </button>
       <button
-        class="btn btn-sm btn-outline-danger mr-3 deleteButton"
+        class="btn btn-sm btn-outline-danger mr-3 delete-button"
         type="button"
       >
         Delete
       </button>
     `
 
-    const startButton = helper.getElement('.startButton', mappingElement)
+    const startButton = helper.getElement('.start-button', mappingElement)
     startButton.onclick = (): void => {
       if (confirm('Are you sure want to start/restart this domain?')) {
         fetch(`/api/mappings/${data.id}/start`)
@@ -134,7 +134,7 @@ class MappingItem {
           )
       }
     }
-    const stopButton = helper.getElement('.stopButton', mappingElement)
+    const stopButton = helper.getElement('.stop-button', mappingElement)
     stopButton.onclick = (): void => {
       if (confirm('Are you sure want to stop this domain?')) {
         fetch(`/api/mappings/${data.id}/stop`)
@@ -146,7 +146,7 @@ class MappingItem {
           )
       }
     }
-    const delButton = helper.getElement('.deleteButton', mappingElement)
+    const delButton = helper.getElement('.delete-button', mappingElement)
     delButton.onclick = (): void => {
       if (confirm('Are you sure want to delete this domain?')) {
         fetch(`/api/mappings/${data.id}`, {
--- a/src/public/manageDomain.ts
+++ b/src/public/manageDomain.ts
@@ -8,6 +8,9 @@ class EnvironmentItem {
   value: string
   isValid = false
 
+  private readonly IS_INVALID = 'is-invalid'
+  private readonly LETTER_NUMBER_REGEX = /^[A-Z0-9_]+$/g
+
   constructor(name?: string, value?: string) {
     this.name = name
     this.value = value
@@ -61,14 +64,14 @@ class EnvironmentItem {
     nameInputElement.addEventListener('input', () => {
       // validate the input for valid variable name with regex
       // valid: letters, numbers and underscore
-      const name = nameInputElement.value.toUpperCase()
-      if (name.match(/^[A-Z0-9_]+$/g)) {
-        this.name = nameInputElement.value.toUpperCase()
+      const upperCaseValue = nameInputElement.value.toUpperCase()
+      if (upperCaseValue.match(this.LETTER_NUMBER_REGEX)) {
+        this.name = upperCaseValue
         this.isValid = true
-        nameInputElement.classList.remove('is-invalid')
+        nameInputElement.classList.remove(this.IS_INVALID)
       } else {
         this.isValid = false
-        nameInputElement.classList.add('is-invalid')
+        nameInputElement.classList.add(this.IS_INVALID)
       }
     })
     valueInputElement.addEventListener(
--- a/src/public/manageDomain.ts
+++ b/src/public/manageDomain.ts
@@ -1,0 +1,149 @@
+const envList = helper.getElement('.envList')
+const environmentVariables: EnvironmentItem[] = []
+const fullDomain = window.location.pathname.split('/').pop()
+document.getElementById('full-domain').innerText = fullDomain
+
+class EnvironmentItem {
+  name: string
+  value: string
+  isValid = false
+
+  constructor(name?: string, value?: string) {
+    this.name = name
+    this.value = value
+    const environmentElement = document.createElement('li')
+    environmentElement.classList.add(
+      'list-group-item',
+      'd-flex',
+      'align-items-center'
+    )
+    envList.appendChild(environmentElement)
+    environmentElement.innerHTML = `
+      <div class="input-group">
+        <input 
+          type="text"
+          aria-label="Name"
+          class="form-control text-uppercase text-monospace"
+          placeholder="NAME"
+          id="name"
+        >
+        <div class="input-group-prepend input-group-append">
+          <span class="input-group-text">=</span>
+        </div>
+        <input
+          type="text"
+          aria-label="Value"
+          class="form-control text-monospace"
+          placeholder="Value"
+          id="value"
+        >
+        <div class="input-group-append">
+          <button
+            class="btn btn-outline-danger"
+            type="button"
+            id="removeVariable"
+          >
+            Remove
+          </button>
+        </div>
+        <div class="invalid-feedback">
+          The variable name must contain only letters, number, and/or the underscore character.
+        </div>
+      </div>
+    `
+
+    const nameInputElement = environmentElement.querySelector(
+      '#name'
+    ) as HTMLInputElement
+    const valueInputElement = environmentElement.querySelector(
+      '#value'
+    ) as HTMLInputElement
+    nameInputElement.addEventListener('input', () => {
+      // validate the input for valid variable name with regex
+      // valid: letters, numbers and underscore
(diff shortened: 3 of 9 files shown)


src/helpers/docker.ts
```diff
@@ -42,7 +51,7 @@ const createContainer = async (
       },
       Tty: false,
       WorkingDir: '/home/node/app',
-      Env: ['NODE_ENV=production', 'PORT=3000'],
+      Env: ['NODE_ENV=production', 'PORT=3000', ...environmentVariables],
```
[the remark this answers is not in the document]
coltonehrman: Is this just a local arbitrary PORT?

src/public/client.ts
```diff
@@ -95,6 +95,13 @@ class MappingItem {
           ${data.gitLink}
         </small>
       </div>
+      <a
+        href="/manage/${data.fullDomain}"
+        class="btn btn-sm btn-outline-primary mr-3 manageButton"
```
coltonehrman, line 100: Just a suggestion, instead of camelCase for CSS classes, maybe use a BEM style? `manageButton => manage-button`

src/public/manageDomain.ts
```diff
@@ -0,0 +1,150 @@
+const envList = helper.getElement('.envList')
```
coltonehrman, line 1: Same suggestion. BEM style class names

src/public/manageDomain.ts
```diff
@@ -0,0 +1,150 @@
...
+  constructor(name?: string, value?: string) {
+    this.name = name
+    this.value = value
+    const environmentElement = document.createElement('li')
+    environmentElement.classList.add(
+      'list-group-item',
+      'd-flex',
+      'align-items-center'
+    )
+    envList.appendChild(environmentElement)
```
coltonehrman, line 20: Should you set the `innerHTML` then `appendChild`? I always imagine "appending/displaying the element into the actual DOM" as the last step.

src/public/manageDomain.ts
```diff
@@ -0,0 +1,150 @@
...
+    environmentElement.innerHTML = `
```
coltonehrman, line 21: Would it be possible to put this HTML template text in perhaps a private const class variable or something out of the constructor? I feel like the constructor is very cluttered.

src/public/manageDomain.ts
```diff
@@ -0,0 +1,150 @@
...
+            Remove
+          </button>
+        </div>
+        <div class="invalid-feedback">
+          The variable name must contain only letters, number, and/or the underscore character.
+        </div>
+      </div>
+    `
+
+    const nameInputElement = environmentElement.querySelector(
```
coltonehrman, line 55: Can use `.getElementByID` API for a bit better performance + better cross-browser compatibility.

src/public/manageDomain.ts
```diff
@@ -0,0 +1,150 @@
...
+const environmentVariables: EnvironmentItem[] = []
+const fullDomain = window.location.pathname.split('/').pop()
+document.getElementById('full-domain').innerText = fullDomain
+
+class EnvironmentItem {
+  name: string
+  value: string
+  isValid = false
+
+  constructor(name?: string, value?: string) {
```
coltonehrman, line 11: I would clean up a lot of the work being done in the constructor by abstracting out into private methods in the class.

src/public/manageDomain.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    const nameInputElement = environmentElement.querySelector(
+      '#name'
+    ) as HTMLInputElement
+    const valueInputElement = environmentElement.querySelector(
+      '#value'
+    ) as HTMLInputElement
+    nameInputElement.addEventListener('input', () => {
+      // validate the input for valid variable name with regex
+      // valid: letters, numbers and underscore
```
coltonehrman, line 63:
```
const upperCaseValue = nameInputElement.value.toUpperCase()
const IS_INVALID = 'is-invalid'
const UNIQUE_REGEX_NAME = /^[A-Z0-9_]+$/g
```

coltonehrman: These are not completely necessary, but it makes it a bit DRYer and more consistent.
You can play with the variable names as well. Doesn't have to be exact.

src/public/manageDomain.ts
```diff
@@ -0,0 +1,150 @@
...
+        this.isValid = false
+        nameInputElement.classList.add('is-invalid')
+      }
+    })
+    valueInputElement.addEventListener(
+      'input',
+      () => (this.value = valueInputElement.value)
+    )
+    environmentElement
+      .querySelector('#removeVariable')
```
coltonehrman, line 79: `.getElementByID()`

src/public/manageDomain.ts
```diff
@@ -0,0 +1,150 @@
...
+      .addEventListener('click', () => {
+        environmentElement.remove()
+        environmentVariables.splice(environmentVariables.indexOf(this), 1)
+      })
+
+    if (this.name && this.value) {
```
coltonehrman, line 85: You can just use `name && value` here.

coltonehrman: Same thing below

src/public/manageDomain.ts
```diff
@@ -0,0 +1,150 @@
...
+      nameInputElement.value = this.name
+      valueInputElement.value = this.value
+      this.isValid = true
+    } else {
+      nameInputElement.classList.add('is-invalid')
```
coltonehrman, line 90: can swap out with the `IS_INVALID`

src/public/manageDomain.ts
```diff
@@ -0,0 +1,150 @@
...
+      '#value'
+    ) as HTMLInputElement
+    nameInputElement.addEventListener('input', () => {
+      // validate the input for valid variable name with regex
+      // valid: letters, numbers and underscore
+      const name = nameInputElement.value.toUpperCase()
+      if (name.match(/^[A-Z0-9_]+$/g)) {
+        this.name = nameInputElement.value.toUpperCase()
+        this.isValid = true
+        nameInputElement.classList.remove('is-invalid')
```
coltonehrman, line 68: can swap out with the `IS_INVALID`

src/public/manageDomain.ts
```diff
@@ -0,0 +1,150 @@
...
+      } else {
+        this.isValid = false
+        nameInputElement.classList.add('is-invalid')
```
coltonehrman, line 71: can swap out with the `IS_INVALID`

src/public/manageDomain.ts
```diff
@@ -0,0 +1,150 @@
...
+      valueInputElement.value = this.value
+      this.isValid = true
+    } else {
+      nameInputElement.classList.add('is-invalid')
+    }
+  }
+}
+
+document.getElementById('addEnvButton').onclick = (): void => {
+  environmentVariables.push(new EnvironmentItem())
```
coltonehrman, line 96: Would be nice to also auto-focus the name input field.

src/public/manageDomain.ts
```diff
@@ -0,0 +1,150 @@
...
+  if (
+    confirm(
+      'Submitting will recreate the container for your app.' +
+        'All data not saved inside the main app folder will be lost. Is this OK?'
+    )
+  ) {
+    // convert to the request body format
+    // { variables: { NAME: value, ... }}
+    const body = environmentVariables.reduce((acc, { name, value }) => {
+      if (name && value) {
```
coltonehrman, line 120: `name && value` should be valid at this point in the code correct? Since you already checked for any invalid fields before?

src/public/manageDomain.ts
```diff
@@ -0,0 +1,150 @@
...
+      return acc
+    }, {})
+    fetch(`/api/mappings/${fullDomain}/environment`, {
+      method: 'PUT',
+      headers: {
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({ variables: body })
+    })
+      .then(response => (response.status !== 201 ? response.json() : {}))
```
coltonehrman, line 132: Can we comment on what `201` status does?

src/public/manageDomain.ts
```diff
@@ -0,0 +1,150 @@
...
+      .then((body: ContainerResponse) =>
+        body.message
```
coltonehrman, line 134: `.message` seems unintuitive as an error. Could we change it to `.error` or even better. Make it actually return with an error status code to be handled with `.catch` or `try...catch`?

coltonehrman (commented):
Can we also see some screenshots with the new UI, please? Thanks 👍

src/public/manageDomain.ts
ggwadera: `message` is consistent with the remaining code base and to the docker API.

src/public/manageDomain.ts
ggwadera: So this was actually a mistake on my behalf. I shouldn't be using IDs here for the buttons since they are not unique in the DOM.

src/public/manageDomain.ts
ggwadera: Also, failed status code are already sent from the server. But since we have to parse the JSON to get the error message, I don't know if there's a simple way to do this inside `.catch()`.

src/public/manageDomain.ts
```diff
@@ -0,0 +1,173 @@
...
+      .addEventListener('click', () => this.removeElement(environmentElement))
+  }
+
+  /**
+   * Sets the input values if they exist. Called on object constructor.
+   */
+  private setInputValues(): void {
+    this.nameInputElement.value = this.name
+    this.valueInputElement.value = this.value
+    this.isValid = true
```
coltonehrman, line 89: I would also probably make a small private function called `setValid` or something.

```
setValid(isValid: boolean): void {
  if (isValid === this.isValid) return
  this.isValid = isValid;
  if (isValid) this.nameInputElement.classList.remove(this.IS_INVALID)
  if (!isValid) this.nameInputElement.classList.add(this.IS_INVALID)
}
```

src/public/manageDomain.ts
```diff
@@ -0,0 +1,173 @@
...
+      </div>
+    </div>
+  `
+
+  constructor(name?: string, value?: string) {
+    this.name = name
+    this.value = value
+    this.createElement()
+    if (name && value) {
+      this.setInputValues()
```
coltonehrman, line 52: I don't really see a need to even call `this.setInputValues` in the constructor, do you?

coltonehrman: You can always set the default values in the constructor and then `setInputValues` should be called upon updates.

coltonehrman: Also, I would probably rename this to something like `updateInputValues`

src/public/manageDomain.ts
coltonehrman: or `updateValidStatus`

src/public/manageDomain.ts
coltonehrman: Unless I'm wrong, I thought any error status codes from `fetch` immediately went to the `.catch()` chain.

src/public/manageDomain.ts
```diff
@@ -0,0 +1,173 @@
...
+  private valueInputElement: HTMLInputElement
+
+  private readonly IS_INVALID = 'is-invalid'
+  private readonly LETTER_NUMBER_REGEX = /^[A-Z0-9_]+$/g
+  private readonly ELEMENT_HTML = `
+    <div class="input-group">
+      <input 
+        type="text"
+        aria-label="Name"
+        class="form-control text-uppercase text-monospace name-input ${this.IS_INVALID}"
```
coltonehrman, line 21: By default, the element is supposed to have `is-invalid` class?

src/public/manageDomain.ts
ggwadera: Only for network errors. From the [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch):

>The Promise returned from fetch() won’t reject on HTTP error status even if the response is an HTTP 404 or 500. Instead, it will resolve normally (with ok status set to false), and it will only reject on network failure or if anything prevented the request from completing.

src/public/manageDomain.ts
```diff
@@ -0,0 +1,173 @@
...
+    environmentElement.innerHTML = this.ELEMENT_HTML
+    environmentElement.classList.add(
+      'list-group-item',
+      'd-flex',
+      'align-items-center'
+    )
+    envList.appendChild(environmentElement)
+    this.nameInputElement = environmentElement.querySelector('.name-input')
+    this.valueInputElement = environmentElement.querySelector('.value-input')
+    this.nameInputElement.addEventListener('input', () =>
```
coltonehrman, line 71: I would probably move the event listeners outside the creat element method. Unless you think it makes more sense here.

```
createAndAppendElements()
setupEventListeners() || handleEvents()
```

coltonehrman: You could also create a private method for each event handler making it even more modular.

src/public/manageDomain.ts
coltonehrman: Interesting ok. I wonder where I got the idea that it would reject lol
Thanks 👍

src/public/manageDomain.ts
ggwadera: It starts empty, and if the value is sent to the constructor then it gets marked as valid. This is to prevent sending variables with no name, which makes no sense.

coltonehrman: I guess it makes sense haha
But, shouldn't the constructor just handle marking it valid or invalid all together?

src/public/manageDomain.ts
ggwadera: Only the name needs some checking when it's input, so a private method makes sense. For the value the event is a simple value assignment, I don't think it needs a specific class method. If this changes in the future it can be refactored.

coltonehrman (approved):
I think it's ready to go now! 👍

songz (approved):

Issue #394, opened 2021-10-17
acme.sh no longer uses let's encrypt by default


songz:
**Describe the bug**
All certificate renewals will fail with latest acme.sh code, because they have moved to the SSL certificate authority to zerossl, which requires registration and has a limit on the number of certs.

To use let's encrypt which our source code is based on, we need to put in the `--server letsencrypt` flag. 

https://github.com/acmesh-official/acme.sh/wiki/Change-default-CA-to-ZeroSSL

**To Reproduce**
Wait for certificates to expire, then the certs may fail one by one.

PR #397, opened 2021-10-29
Updates domainSetup with letsencrypt

mino323:
Adds let encrypt flag so it doesn't need to be changed during setup.


songz:
closes #397 https://github.com/garageScript/myProxy/issues/394

songz (approved):

PR #399, opened 2021-10-31
feature: resolve #396 add git to containers running apps

### LINKED ISSUE #396: Feature: Add git to Docker Container
If the web application deployed needs to use `git` then the application would error out since the `node:alpine` image used by myProxy does not have git installed.

hwong0305:
Code updates containers to be created from devwong0305/myproxy-node.

Which is a node:alpine with git installed.

### DIFF
--- a/scripts/Dockerfile
+++ b/scripts/Dockerfile
@@ -1,0 +1,8 @@
+FROM node:alpine
+
+RUN apk add git --no-cache
+RUN git config --global user.email "myproxy@garagescript.org"
+RUN git config --global user.name "myproxy"
+
+ENTRYPOINT ["docker-entrypoint.sh"]
+CMD ["node"]
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -91,7 +91,7 @@ fi
 
 # pull node docker image
 if docker ps > /dev/null 2>&1; then
-  docker pull node:alpine
+  docker pull devwong0305/myproxy-node:latest
 else
   echo "WARNING: Couldn't run docker commands"
   echo "WARNING: Make sure your user has the right permissions"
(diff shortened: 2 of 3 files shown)


tomrule007:
Would it be possible to just include the container configuration file and build step in the setup file? That way it is more transparent to users of myproxy what is happening

ggwadera:
> Would it be possible to just include the container configuration file and build step in the setup file? That way it is more transparent to users of myproxy what is happening

This could be done by including a dockerfile and building the image locally when the setup script is executed. This also opens the possibility for more customizations if desired.

hwong0305:
I actually just used a Dockerfile to build the new image. I am not sure we want to open this so users can upload their own dockerfile.

Unfortunately, I did look at the API, and I didn't see the way to add git without creating a new image.

ggwadera:
> I actually just used a Dockerfile to build the new image. I am not sure we want to open this so users can upload their own dockerfile.
> 
> Unfortunately, I did look at the API, and I didn't see the way to add git without creating a new image.

What I meant is that instead of pulling the container image, you include the dockerfile you used to create it, so the image is built when the setup.sh script is executed.

This wouldn't allow every user to change the image, since it would be built only when setting up the server. And it's also not a feature to allow arbitrary dockerfiles to be uploaded.

hwong0305:
> > I actually just used a Dockerfile to build the new image. I am not sure we want to open this so users can upload their own dockerfile.
> > Unfortunately, I did look at the API, and I didn't see the way to add git without creating a new image.
> 
> What I meant is that instead of pulling the container image, you include the dockerfile you used to create it, so the image is built when the setup.sh script is executed.
> 
> This wouldn't allow every user to change the image, since it would be built only when setting up the server. And it's also not a feature to allow arbitrary dockerfiles to be uploaded.

PR updated.

ggwadera (approved):

Issue #390, opened 2021-06-04
Possible source of ssh key issue.


tomrule007:
I believe this if statement is always failing because of the quotes around the file name. I'm still a noob with bash so i very well could be missing something but thought i would pass it along.

tested in the terminal with:
```
[ -f  "~/.ssh/authorized_keys" ] || echo "fail"
[ -f ~/.ssh/authorized_keys ] && echo "success"
```

https://github.com/garageScript/myProxy/blob/022ace46b2014e7a8ebdb89c14538b11a911f09f/scripts/setup.sh#L64-L72

ggwadera:
Nice catch, I thinkg the `~` is not expanded because of the quotes, should've been `[ -f "$HOME/.ssh/authorized_keys" ]` to expand into the correct path.

Either way this might not be the issue since this script only runs when first starting the server. After that the process is handled by PM2.

https://github.com/garageScript/myProxy/blob/022ace46b2014e7a8ebdb89c14538b11a911f09f/package.json#L20

SlyBouhafs:
Possible fix with #406

PR #406, opened 2022-07-29
Update ssh keys path to the right path 

SlyBouhafs:
This PR will fix the path that is used by myproxy at startup so that it it gets all the old keys right.

Here, we can see the path that gets updated with the new keys is indeed `/home/git/.ssh/authorized_keys`: https://github.com/garageScript/myProxy/blob/ce94f9976d57ce3b0faaabfb79877abc7dd025ba/src/helpers/authorizedKeys.ts#L11

### DIFF
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -37,7 +37,7 @@ const startAppServer = (
     setPass(adminPass)
 
     if (isProduction()) {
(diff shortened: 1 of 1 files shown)


flacial (approved):

SlyBouhafs:
Possible fix for #390

